PR #13, opened 2019-08-28
fix: restore test command and deal with standard deps

amcgee:


cli/config/jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
varl: We may need to set up something for styled-jsx for use with Jest.

cli/config/standard-libs.json
```diff
@@ -0,0 +1,11 @@
+{
```
varl: Any idea if we can get dependabot to look at this file as dependencies?

cli/package.json
```diff
@@ -15,21 +15,31 @@
         "src"
     ],
     "dependencies": {
+        "//": "****** Includes standard libraries from app-shell, so that tests work ******",
```
varl: This is needed for the same reason as we use `dependencies` in `cli-style`, right? To get them installed when a package installs this as a dep?

cli/config/jest.config.js
amcgee: What do you mean by this?

cli/config/standard-libs.json
amcgee: Probably if we made it a package.json, but only the names are actually used at present.  I tried to consolidate the standard deps into a common package so we didn’t commingle them or duplicate them (4 places currently).  I couldn’t get it to work nicely though, so left them as duplicates for now

cli/package.json
amcgee: Yeah - they’re not needed for production or development builds which get injected into the she’ll, but when running local tests they’re run outside the shell so they need to be able to find these peerdeps somewhere somewhere

cli/config/standard-libs.json
varl: Alright, problems for a bright future to solve.

cli/config/jest.config.js
varl: https://github.com/zeit/styled-jsx#styled-jsxcss-in-tests

I think @HendrikThePendric and @jenniferarnesen ran into problems using styled-jsx components with Jest and had to do some configuration. I don't remember the details. Could deal with that when we get there.

amcgee: Got it.  Tests do seem to work with `styled-jsx` at least with in-line styles - the example installed by `init` has `<style jsx>{...}</style>` and renders properly.  I pass our custom `babel` config explicitly to `babel-jest` which includes the `styled-jsx` transpiler, so maybe that's all that was needed?

amcgee: Actually looks like it's an issue with tagged template literals, which aren't working (again) anyway.  Will need to look into that shortly.

Issue #26, opened 2019-09-02
app-scripts build ignores package.json's 'homepage' field


awgaan:
When building a production build the _homepage_ field in _package.json_ is ignored.

## Expected Behavior
With
```json
homepage: "abc"
```
```
The project was built assuming it is hosted at abc/.
You can control this with the homepage field in your package.json.
```

## Current Behavior
With
```json
homepage: "abc"
```
```
The project was built assuming it is hosted at the server root.
You can control this with the homepage field in your package.json.
```

## Steps to Reproduce
1. Set a value for homepage key in package.json
2. yarn run d2-app-scripts build

## Context (Environment)
```json
    "@dhis2/cli-app-scripts": "^1.2.0"
    "@dhis2/app-runtime": "^1.5.1",
```

amcgee:
See comment on #27 - thanks for testing!

PR #27, opened 2019-09-02
fix: added option to override homepage

awgaan:
See #26 

Maybe it would be better to define a default value for the homepage so that if neither _package.json_ or _d2.config.js_ has that field we can use a default value instead of leaving it as undefined (which works since _react-scripts build_ ignores the value if undefined.

### DIFF
--- a/cli/src/lib/parseConfig.js
+++ b/cli/src/lib/parseConfig.js
@@ -52,6 +52,7 @@ const parseConfig = paths => {
 
         if (fs.existsSync(paths.package)) {
             config.name = config.name || require(paths.package).name
+            config.homepage = config.homepage || require(paths.package).homepage
         }
         config.title = config.title || config.name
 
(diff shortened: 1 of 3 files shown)


amcgee:
Thanks for your contribution @awgaan !  This is really important.

I'm actually going to change the implementation somewhat so that the produced `homepage` CRA setting is automatically generated to be "correct by default", but can be overrided with the PUBLIC_URL environment variable as you did here - we can generate the manifest and the default PUBLIC_URL and the server rootUrl for a production build automatically from just the application name!

amcgee:
See #36 and let me know what you think - it's a bit of a bigger change than this one but I think it's absolutely the "right thing to do" to immediately create a zip file that can be uploaded to a DHIS2 instance

awgaan:
@amcgee That's perfect. Thank you for prioritizing this issue. Makes it much easier for us to use the app-platform in the course. Let me know if I can help on something else; be it testing, documentation or whatever.

PR #34, opened 2019-09-05
fix: improve start behavior and logging, don't orphan react-scripts, add postcss

amcgee:
This has one important functional change - we wait for the app build to succeed before starting the appShell server, so we don't run the risk of orphaning the server process if/when the first compilation fails (a fatal error in rollup).  This also:

- Adds support for overriding the server port with the PORT env var
- Significantly improves the aesthetic appeal and consistency of console output
- Stops automatically piping the react-script output

### DIFF
--- a/cli/config/rollup.config.js
+++ b/cli/config/rollup.config.js
@@ -6,6 +6,7 @@ const resolve = require('rollup-plugin-node-resolve')
 const babel = require('rollup-plugin-babel')
 const commonjs = require('rollup-plugin-commonjs')
 const json = require('rollup-plugin-json')
+const postcss = require('rollup-plugin-postcss')
 
 const { reporter } = require('@dhis2/cli-helpers-engine')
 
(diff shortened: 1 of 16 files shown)


amcgee:
Here's an example of what the output looks like when running `start` and encountering a few errors:
Screenshot 2019-09-05 17 35 01

varl:
The addition of `postcss` support makes this a feature bump.

I don't have the opportunity to try this tonight but will give it a whirl in the morning. Looks pretty slick!

amcgee:
I was thinking about dropping postcss and adding it later, but figured since the compilation failed when it hit a CSS import before it could be considered a fix (doesn't change the API).  Either way fine with me, we might want to do CSS imports differently in the future anyways.

The library in question that @erikarenhill was trying to use was @dhis2/analytics - this gets past the CSS import issue but there are other issues we'll need to deal with as well.

Taking out PostCSS for now

varl:
Yeah, avoiding the css-loader situation was one of the reasons we rewrote ui-core from postcss to styled-jsx.

amcgee:
> Yeah, avoiding the css-loader situation was one of the reasons we rewrote ui-core from postcss to styled-jsx.

Yep, agreed.  Libraries with `styled-jsx` should work, but we'll probably also have to at least temporarily support direct CSS imports (most notably for `d2-ui`)

I'll open another PR next week to address this and other issues importing some of our legacy libraries

PR #47, opened 2019-09-17
fix: remove publish-breaking comment from package.json

amcgee:
Publishes have been silently failing (see dhis2/cli#131) since 1.2.0 because of the comment dependency in package.json.  I moved the comment out of the dependencies map, which should work fine.


amcgee:
Going to go ahead and merge this to kick off the deploy

PR #52, opened 2019-09-20
feat: improve test command, support dotenv files, add postcss

amcgee:
This is an upgrade to the test command which adds support for jest config overrides, provides nice testing defaults, and adds support for `--updateSnapshot`, `--coverage`, and `--watch` to the `test` command.

This also adds support for specifying environment variables in `.env` files (including `.env.local`, `.env.<mode>[.local]` (**mode** == _production_, _development_, or _test_), for all three build/test/start commands

This does modify the behavior of the compiler somewhat - we now only bundle **apps** into a single file, not **libs** (this might change again in the future).  We also only support `CJS` module loading (for now) because of an issue with CJS named exports in static dependendency resolution (see [the code comment](https://github.com/dhis2/app-platform/pull/52/files#diff-d6f0407554b6e696594d77ffd980e00fR77) and [the rollup issue](https://github.com/rollup/rollup-plugin-commonjs/issues/211#issuecomment-337897124))

This also adds support for `postcss` so that we can more easily port CRA apps to the platform.  It supports `import 'my.css'` for global styles, as well as `import styles from 'my.module.css'` for CSS module support (filenames must have the extension `.module.css`)

Finally, I've patched up a few small issues in the shell, such as an omitted typeface-robotto and including more favicon variants

These changes were influenced by the effort to port Usage Analytics to the platform, see dhis2/usage-analytics-app#94


shell/public/index.html
```diff
@@ -1,38 +1,69 @@
...
-    <!--
-      manifest.json provides metadata used when your web app is installed on a
-      user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
-    -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
-    <!--
-      Notice the use of %PUBLIC_URL% in the tags above.
-      It will be replaced with the URL of the `public` folder during the build.
-      Only files inside the `public` folder can be referenced from the HTML.
+<html lang="en" dir="ltr">
```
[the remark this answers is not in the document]
amcgee: I copied most of this from the CRA index.html in Usage Analytics, but that's probably actually an older version of CRA now.  Will re-add it.

amcgee:
> I'm a bit way about misuse of .env files, though that is up to the consumer. The 12-factor app principles don't always apply to us, nor should they, though some of the points are general enough to apply where we think they make sense.

I think I agree here - it's a tradeoff, though, since `.env` files I think can make development (and particularly shared development) easier.  I went for CRA compatibility here, to minimize the modifications necessary to i.e. `usage-analytics-app`.

I'm onboard with removing at least the per-environment env files, but would be curious what the "right" way to implement [this pattern](https://github.com/dhis2/usage-analytics-app/blob/master/.env.production) should be in the platform - I know we'll be getting rid of this use of source-local HTTP requests soon, so we could omit the `REACT_APP` variables, but there will be different configurations of `DHIS2_BASE_URL` between development and production at least for the foreseeable future.

@varl what do you envision - no support for dotenv or support for only *one* dotenv (possibly with `.local` overrides)

varl:
We need to think a bit about it.  For instance, what is an environment? My computer for example is a development environment, and therefore it should not be allowed to act as a production environment.

Generally I think that a `.env.template` with all variables committed to the repository, and an ignored `.env` file is enough. The `.env` file has the environment variables for the current environment. That's it. This doesn't really work for us since we cannot e.g. have a `.env` file on Travis for CI, or one on Netlify or one on the production instance that can be read by the app. 

CRA does not use the environment files for _environments_ in the strictest sense, it uses it for configuration for different modes for the intended environment, which solves the problems above in exchange for a bit of a mess of files in the app repo.

Anyway, pardon me for bringing up this topic in this PR. I think this is good for now.

amcgee:
@varl it's a very fair point since this adds support for `.env[.mode][.local]` files, should we pull that support and just allow `.env` for now or merge this with support for all variants?

amcgee:
The advantage of allowing multiple mode-targeted `.env` files is that you can, for instance, specify `http://my-server.com` for development and then revert to `../../../` (or whatever) in production.  We could set this automatically, but I also think there might be situations where a production build would want a custom root url

amcgee:
(p.s. thanks for the feedback!)

varl:
> @varl it's a very fair point since this adds support for `.env[.mode][.local]` files, should we pull that support and just allow `.env` for now or merge this with support for all variants?

I thought about that too, and to be pragmatic about it; `.env[.mode]` makes a lot of sense for configuration purposes. The case for `.env[.mode][.local]` is weaker as `.env.local` is covers the most reasonable use cases.

In the end, having support for it doesn't mean we _have_ to use it, so we can safely leave it in.

shell/public/manifest.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "Usage Analytics App",
+    "short_name": "UAA",
+    "description": "Gain insights into the usage of your DHIS2 instance",
```
HendrikThePendric: I think these hardcoded texts about the Usage Analytics App need to be replaced by something dynamic. Perhaps using info from `package.json`

varl: Hehe. Good catch.

ismay: We could also remove it. I wonder how many users install our apps on their home screen, and we're not that focused on building PWA's either. Other than that I don't know if there's any benefit: https://developers.google.com/web/fundamentals/web-app-manifest

ismay:
> In the end, having support for it doesn't mean we have to use it, so we can safely leave it in.

Nice. I personally like using this to have a standard way to point to a dhis2 backend. I don't like the approach we used before with the dhis2 config that lives outside of the repo, on your system somewhere. The .env files make way more sense to me.

shell/public/manifest.json
varl: One of our teams has built a prototype converting the Dashboards app to a PWA with pretty good success. We might end up using that style for some apps that make sense. We probably do not want people to run Maintenance app on their phones, but for a tablet light maintenance or viewing visualisations can make sense.

And, for external developers it is nice if we have built-in support for PWAs, so I do think having the option is nice.

amcgee: Good catch @HendrikThePendric 

I actually thought I had removed this file (I removed the link from index.html), which is why the UA info snuck in there.

I agree we should think about supporting mobile and PWA apps in the future, but I think I'd advocate for leaving out `manifest.json` until we have a solid strategy around that.  Additionally, we would need to create a whole templating system for static files in order to make the manifest.json data custom - CRA only supports %VAR% substitutions in `index.html`, so I don't think that's worth it at this point.  We currently generate `webapp.manifest` at build time, so we could in the future extend that to create `manifest.json` as well.

Anybody opposed to just leaving out `manifest.json` for now and adding it later?

ismay: > And, for external developers it is nice if we have built-in support for PWAs, so I do think having the option is nice.

Yeah that sounds like a good reason.

varl: @amcgee, if it's in the name of scoping down: :+1:

amcgee: Doing it

amcgee:
> I've left one comment in the code that I think needs to be addressed before this is merged.

@HendrikThePendric removed the offending file, thanks for catching that - should be good to go now!

amcgee:
I'd really like to get this in - @HendrikThePendric I think your request has been addressed, let me know ASAP if it hasn't, otherwise I'd like to merge this first-thing tomorrow!

amcgee:
Thanks all for the feedback!

PR #97, opened 2019-10-02
fix: don't dynamically load the app adapter (prevent blank flash)

amcgee:
I switched the AppAdapter to a React.Lazy component a while back.  This has the advantage of code-splitting the bundle, so that we can analyze what's in the adapter vs the shell itself.  However, it had the consequence of causing the Headerbar to wait until the entire application has loaded before rendering.  This meant that, for applications with moderate bundle sizes like the Query Playground and also in low-bandwidth situations the entire page would flash "blank white" for a noticeable duration.

This change reverts the Adapter code-splitting, so that the shell displays the adapter (including the headerbar) on first render.  It also adds displays full-page screen cover and loading spinner while the app bundle loads, rather than an empty div.  You can test this behavior by setting "slow 3g" in dev tools.  Note that there is also a loading spinner displayed while the AuthBoundary loads its data - we'll want to consolidate these once we have either first-class suspense for data loading (from React) or implement a LoadingBoundary ourselves (using Context)

### DIFF
--- a/examples/simple-app/yarn.lock
+++ b/examples/simple-app/yarn.lock
@@ -36,17 +36,17 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.6.0":
-  version "7.6.0"
-  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.6.0.tgz#9b00f73554edd67bebc86df8303ef678be3d7b48"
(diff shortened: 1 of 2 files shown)


varl:
Other than the `yarn.lock` change this is for the best. Having the header bar visual early is a nice trait.

amcgee:
@varl the yarn.lock changes are in `examples/simple-app` and come from [this line](https://github.com/dhis2/app-platform/pull/97/files#diff-da17543c5fcabc48ca692ec1a856dc9aR930)

Basically because the dep is a **file** dependency, the yarn.lock changes there are delayed until after a release (version bump)

varl:
Interesting. Thanks!

PR #129, opened 2019-10-21
fix: use browser field in package.json if it exists

amcgee:
We should use the `browser` field in package.json for dependency resolution, otherwise we might get server code (for packages that distribute both under the same name, like `styled-components`)

> NOTE: The yarn.lock changes in the example app and shell adapter are unrelated, just residual from the previous version bump


varl:
Perfect. Thanks for the heads up on the lock-bump too. 👍

PR #162, opened 2019-11-12
fix: upgrade Headerbar to 2.0.1, don't break on older servers

amcgee:
Also fixes the build process to ensure `shell/yarn.lock` is correctly updated


package.json
```diff
@@ -19,10 +19,11 @@
     "scripts": {
         "format": "d2-style js apply --all --no-stage",
         "install:example": "cd examples/simple-app && yarn install --force --check-files",
+        "install:shell": "cd shell && yarn",
```
varl: Neat!

Issue #125, opened 2019-10-16
Add tests running `init` command and verifying working shell with cypress


amcgee:

amcgee:
Ping

varl:
Superseded by: https://jira.dhis2.org/browse/LIBS-29

PR #198, opened 2019-12-06
chore(deps): upgrade @dhis2/ui-core to 4.1.1 and @dhis2/ui-widgets to 2.0.4

edoardo:
This is needed for DV where the new options UI uses features and fixes in `@dhis2/ui-core` introduced from version 4.

BREAKING CHANGE: This will break applications expecting the `ui-core` v3 API

### DIFF
--- a/shell/yarn.lock
+++ b/shell/yarn.lock
@@ -931,8 +931,7 @@
 "@dhis2/app-adapter@file:./adapter":
   version "1.0.0"
   dependencies:
-    "@dhis2/ui-core" "^3.12.0"
-    "@dhis2/ui-widgets" "^2.0.2"
+    "@dhis2/ui-widgets" "^2.0.4"
     moment "^2.24.0"
 
 "@dhis2/app-runtime@^2.0.4":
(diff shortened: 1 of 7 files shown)


varl:
This requires changes in:

https://github.com/dhis2/app-platform/blob/master/shell/adapter/src/AuthBoundary/LoginModal.js

the `onChange` handlers now have the signature `object, event`.

edoardo:
> This requires changes in:
> 
> https://github.com/dhis2/app-platform/blob/master/shell/adapter/src/AuthBoundary/LoginModal.js
> 
> the `onChange` handlers now have the signature `object, event`.

Good catch.
Fixed.

amcgee:
I think we currently have two copies of `ui-core` in the shell, since our version of `ui-widgets` uses `ui-core@4.x`?  We should think about how to consistently upgrade these (especially with breaking changes, since these will break apps) in general.

We probably don't want to wait for it, but this would be a good use-case for #125 - it would probably have caught the Login Dialog failure due to `ui-core` API changes

shell/yarn.lock
```diff
@@ -955,7 +955,14 @@
...
 
-"@dhis2/ui-core@3.12.0", "@dhis2/ui-core@^3.11.0", "@dhis2/ui-core@^3.12.0":
+"@dhis2/prop-types@^1.4.0":
+  version "1.5.0"
+  resolved "https://registry.yarnpkg.com/@dhis2/prop-types/-/prop-types-1.5.0.tgz#7e69919f66698be373dd21940a8a770234ded6a1"
+  integrity sha512-dueFkkAMOIMbXiU7Mhr3Y+DBRyOd/rHA+5/IDiYWN1xttlUTSuGZLQ5AnJ7osBicEhx+qElaGbTdRYQj3SMBtA==
+  dependencies:
+    prop-types "^15"
+
+"@dhis2/ui-core@3.12.0", "@dhis2/ui-core@^3.12.0":
```
amcgee: Where are these versions of `ui-core` coming from?

amcgee:
> Where are these versions of ui-core coming from?

This was from an older version of `ui-widgets` which bundled its own `ui-core` - fixed in 70d4f8499db2db3eab3a6a635ef9762215252872

amcgee:
This was actually broken because `LoginDialog` still referenced the non-existant `Modal.Title`, `Modal.Content`, and `Modal.Actions` components - I had to switch them here a00d647

@edoardo probably good to get in the habit of running the example app (`yarn start` in root directory) to sanity-check breaking dependencies, especially until we have #125 

I also added `BREAKING CHANGE` to the issue description, we need to make sure this is included in the commit **footer** when merging.

PR #221, opened 2019-12-17
feat: publish shell and adapter, fix dependency resolution

### LINKED ISSUE #196: Application dependency resolution doesn't work well
Application dependencies aren't deduplicated with shell dependencies since #187, which can lead to duplicates in the bundle and React Hooks issues.

(credit @edoardo for pointing this out)

amcgee:
Fixes #196 

Use npm to install `app-shell` and `app-adapter` from the registry, rather than bundling them as assets.  This allows us to resolve all dependencies through the top-level application's `node_modules`, rather than through the subsidiary shell's `node_modules`.  This also allows shell dependencies to be overridden with the `resolutions` field in the app's `package.json`, though that's not generally recommended.

To make this even cleaner, we should show a warning if the hoisted version of an app-shell dependency doesn't match the version resolved from the shell itself, and also detect when multiple copies of dependencies (particularly singleton ones like `react` and `styled-jsx`) end up in the bundle.


amcgee:
Thanks @varl for the idea to try this approach and the productive brainstorming!

varl:
This looks pretty clean, @amcgee. Good work!

PR #222, opened 2019-12-18
fix: add an eslint config to shell, falling back to react-app defaults

### LINKED ISSUE #197: ESLint config not found when missing from application root
And eslint plugin dependency resolution gets funky when adding a `.eslintrc.js` to `/shell`.

(Credit @awgaan)

amcgee:
Fixes #197

This adds an eslint config to the shell root which will automatically detect a top-level `.eslintrc.js` file and load it, or will use `react-app` as a fallback.

### DIFF
--- a/shell/.eslintrc.js
+++ b/shell/.eslintrc.js
@@ -1,0 +1,13 @@
+const fs = require('fs'),
+    path = require('path')
+
+const customConfigFile = path.resolve('../../.eslintrc.js')
+const hasCustomConfig = fs.existsSync(customConfigFile)
+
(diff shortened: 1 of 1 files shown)


amcgee:
We could consider automatically enforcing our cli-style React ESLint config, but I think this is OK for now.

amcgee:
> Should work fine since `react-app` plugin for ESLint should be available with the shell (through CRA), right?

[Yep, it's a dep of `react-scripts`](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/package.json#L46)

PR #265, opened 2020-01-15
feat: allow customization of the dhis2 app icon

### LINKED ISSUE #212: Support custom app icons
Currently the application icon (favicon and also, more importantly, icon in app store and app menu) is the DHIS2 logo for all platform apps - this should be overridable so the application can specify its own icon (without hacks)

amcgee:
There are certainly more versatile ways to do this, but this seems like a preetty simple fix for the moment.  I'm very open to counter-suggestions (specifying any file in `d2.config.js`, for instance) or recommendations on another filename (currently `public/dhis2-app-icon.png`).  We could also not provide a default and **require** that the app developer specifies one, but that seems harsh for the moment.

For now the favorite icon will remain the DHIS2 logo (though undocumented, this can also be overridden in the same way, you just have to do it for a lot of different file sizes).

Fixes #212


docs/usage/app-icon.md
```diff
@@ -0,0 +1,7 @@
+# App Icon
+
+The App Icon should allow a user to quickly and uniquely identify your application. It will appear in the DHIS2 App Hub and in the list of available applications in the Headerbar of a running DHIS2 instance.
+
+## Specifying a Custom App Icon
+
+To specify a custom App Icon, place the relevant file in the `public` directory and name it `dhis2-app-icon.png`. It must be in PNG format, and should include a transparent background. The recommended size of this image is 48x48 pixels. When no `public/dhis2-app-icon.png` is provided, a default icon (the DHIS2 logo) will be used.
```
varl: I'd remove this: "(the DHIS2 logo)" so we can just change the default icon without having to remember to update the docs.

amcgee: Will do

PR #267, opened 2020-01-16
fix: control the viewport, always show the headerbar, size app container

amcgee:
This ensures that the appShell takes full control of the browser viewport, that the Headerbar is "sticky" (actually never moves), and that the app has a sized container so that it can properly render full-height divs.

This is required for the DV port, since it currently can't know how high the viewport is without a hard-coded Headerbar height


amcgee:
See https://github.com/dhis2/data-visualizer-app/issues/494

varl:
Do you mean that the headerbar should always be visible at the top of the screen, even when the user scrolls down on the page?

amcgee:
> Do you mean that the headerbar should always be visible at the top of the screen, even when the user scrolls down on the page?

Yeah, @cooper-joe species the headerbar to be always visible.  Instead of scrolling the whole page you scroll in the content div.  we could also affix the headerbar above the content but I think I like this better (and easier because you don’t need a matching-height buffer).

varl:
@amcgee this could be merged as a feature depending if "sticky headerbar" is considered an addition or a bug.

amcgee:
It could, not really sure.  The lack of viewport sizing was definitely a big.

varl:
Yeah, ideally we'd land two commits on `master`: one `fix` and one `feat`.

Since this isn't a feature that a user can opt in to use, considering and landing it as a single `fix` is fine.

PR #302, opened 2020-02-18
fix(cli-app-scripts): use the same build dir for modes

varl:
This removes the devOut path and usage of it, instead opting for
building the development build and production build into the same
folder: ./build.

For libraries in monorepo situations this is important as they will look
up the main or module key in the package.json file, which is usually set
to e.g. ./build/{cjs,es}/lib.js when using the cli-app-scripts to build
the lib.

### DIFF
--- a/cli/src/lib/compiler/compileLibrary.js
+++ b/cli/src/lib/compiler/compileLibrary.js
@@ -24,10 +24,7 @@ const compileLibrary = async ({ config, paths, mode, watch }) => {
     const input =
         (config.entryPoints && config.entryPoints[config.type]) ||
         'src/index.js'
-    const outDir = mode === 'production' ? paths.buildOutput : paths.devOut
-
-    fs.removeSync(outDir)
-    fs.ensureDirSync(outDir)
+    const outDir = paths.buildOutput
 
     const pkg = require(paths.package)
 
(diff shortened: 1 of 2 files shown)


cli/src/lib/compiler/compileLibrary.js
```diff
@@ -81,6 +81,8 @@ const compileLibrary = async ({ config, paths, mode, watch }) => {
             process.exit(1)
         }
     } else {
+        fs.ensureDirSync(outDir)
```
amcgee: Is there a reason not to remove the existing directory before running the compiler in watch mode?

varl: Yes, Webpack Dev Server crashes if the directory disappears and needs to be restarted. Just overwriting the files works however, but we need to ensure that the dir is there on first run.

amcgee: Meaning webpack devserver somewhere else in the monorepo?  Unfortunate, but OK for now 👌

varl: Yeah, for example when I run Storybook for ui-core and the `build` folder disappears, Storybook crashes, so deleting it on every rebuild is a bit annoying.

PR #325, opened 2020-03-20
fix: allow login username or password to be shorter than 4 chars

amcgee:
On the DHIS2 user registration page we require a minimum of 4 characters when selecting a username.  However, in the user management app the minimum is 2 characters.

Screen Shot 2020-03-20 at 10 29 12 PM

This change allows users created in the latter interface to log in through the AppShell login dialog.


amcgee:
Big thanks to @chathurawidanage for picking up on this inconsistency!!

PR #332, opened 2020-03-25
fix: don't generate styled-jsx randomized classes in jest snapshots

amcgee:
In test environments we should:

1. Use `styled-jsx/babel-test` instead of `styled-jsx/babel`
2. Mock the `styled-jsx/css` function as a no-op

This results in clean snapshot output without randomly-generated `jsx-#####` which change whenever any css changes.  See [the styled-jsx README](https://github.com/zeit/styled-jsx/blob/master/readme.md#rendering-in-tests)

### DIFF
--- a/cli/config/app.babel.config.js
+++ b/cli/config/app.babel.config.js
@@ -24,8 +24,22 @@ module.exports = {
 
         // Automatically include a React import when JSX is present
         require('babel-plugin-react-require'),
-
-        // Always build in "production" mode even when styled-jsx runtime may select "development"
-        [require('styled-jsx/babel'), { optimizeForSpeed: true }],
     ],
+    env: {
+        production: {
+            plugins: [
+                [require('styled-jsx/babel'), { optimizeForSpeed: true }],
+            ]
+        },
+        development: {
+            plugins: [
+                [require('styled-jsx/babel'), { optimizeForSpeed: true }],
+            ]
+        },
+        test: {
+            plugins: [
+                require('styled-jsx/babel-test')
+            ]
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


cli/config/jest.styled-jsx-css.mock.js
```diff
@@ -0,0 +1,5 @@
+function css() {
+    return ''
+}
+
+module.exports = css
```
HendrikThePendric: I reckon we'll also need to moch `resolve`

HendrikThePendric: The article doesn't mention this specifically, so perhaps it's not required... But we do tend to use `resolve` and I think it probably needs the same treatment as `css`.

amcgee: Thanks @HendrikThePendric, good catch - I added both `resolve` and `global` functions and also mocked the `macro` entrypoint in 2fac1ce

cli/config/app.babel.config.js
```diff
@@ -24,8 +24,22 @@ module.exports = {
...
     ],
+    env: {
+        production: {
+            plugins: [
+                [require('styled-jsx/babel'), { optimizeForSpeed: true }],
+            ]
+        },
+        development: {
+            plugins: [
+                [require('styled-jsx/babel'), { optimizeForSpeed: true }],
```
amcgee: @varl should we turn `optimizeForSpeed` off here?

amcgee: Note this is for both apps and libraries, but I think it should be safe to build without `optimizeForSpeed` when the build output is in development...?

amcgee: Anyway, will merge this and we can consider later.

varl: The tricky thing is when production-mode components and development-mode components interact on the same page. Since styled-jsx is a singleton, that causes problems as production/development components inject CSS differently and are not compatible.

Issue #353, opened 2020-04-08
Documentation improvement on App Platform Scripts page 


paschalidi:
Hello y'all 👋 

I was going through the documentation and I found that some part of the App Platform Scripts doesn't make much sense for me. I am bringing it up so that it can be discussed if it is only me that had a little bit of trouble to understand what is going on there.

Look [here](https://github.com/dhis2/app-platform/blame/3e769977151cd48cd787b697c309e26c45930868/docs/scripts.md#L7) for the code and [here](https://platform.dhis2.nu/#/scripts) for the website. 

My suggestion would be that we explicitly make clear that the `init` command will install certain packages and one of those would be `d2-app-scripts` however if the user wants to install the package as a standalone they can do by running `yarn add --dev @dhis2/cli-app-scripts`. 

It is a very simple issue but I hope that it makes sense that we discuss it first before I open a PR. Le me know what you think and I will be happy to go on and rephrase this.

amcgee:
Hi @paschalidi thanks for the issue - did you see the earlier documentation page `installation` which lays out these two options pretty explicitly?  If you have recommendations for how to clarify the other documentation pages, feel free to make a pull request!

varl:
Superseded by: https://jira.dhis2.org/browse/LIBS-21

PR #362, opened 2020-04-14
docs: rephrases already existing documentation

paschalidi:
Hey @amcgee 👋 

This is what I meant with my issue here: https://github.com/dhis2/app-platform/issues/353

Have a look and feel more than free to close the issue if it doesn't make sense to you :)

### DIFF
--- a/docs/scripts.md
+++ b/docs/scripts.md
@@ -4,4 +4,4 @@ The application platform comes with a set of scripts. This tool is packages as `
 
 To run the `init` script as a stand-alone tool, either install the D2 cli globally (`yarn global add @dhis2/cli`) and run `d2 app scripts init myapp`, or run the `init` script by itself with `npx` (`npx @dhis2/cli-app-scripts init myapp`)
 
-To install `d2-app-scripts` as a dev dependency (this is done for you by the `init` command) run `yarn add --dev @dhis2/cli-app-scripts`.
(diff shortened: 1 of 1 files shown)


amcgee:
Hey @paschalidi, thanks for the suggestion!  For me this is not as clear as the original, but I may have some bias - @dhis2/team-apps anyone have an opinion here?  (low-priority)

paschalidi:
As you said is simply a suggestion.

Mohammer5:
I'm in favor of keeping the current version as it highlights the distinction between manual and automated installation of the dev dependency while the change tries to highlight a distinction between a globally and locally available package, which should be an addition the the former instead of replacing it if we think we need this distinction.

amcgee:
Thanks @Mohammer5, well-put!  @paschalidi do you want to update this so that it adds a new sentence about "global" installation and usage?

amcgee:
Actually the preceding paragraph does this already:

> To run the `init` script as a stand-alone tool, either install the D2 cli globally (`yarn global add @dhis2/cli`) and run `d2 app scripts init myapp`, or run the `init` script by itself with `npx` (`npx @dhis2/cli-app-scripts init myapp`)

Feel free to suggest a clarification there if needed, otherwise I think we can close this PR?

paschalidi:
Yes we can close the PR indeed. Thanks both for looking into it!!

PR #405, opened 2020-05-20
fix: add login form data-test props

Mohammer5:
This will add unique `data-test` props to the login form so that logging in can easily be automated in in-browser tests.


amcgee:
@Mohammer5 Should we make this a `fix` so it triggers a release though?

PR #416, opened 2020-06-09
chore: replace ui-core & ui-widgets v4 with ui v5

HendrikThePendric:
BREAKING CHANGE: components should now be imported from `@dhis2/ui` and various components have had breaking changes in v5


adapter/package.json
```diff
@@ -29,7 +28,7 @@
     "peerDependencies": {
         "@dhis2/app-runtime": "^2.0.4",
         "@dhis2/d2-i18n": "^1.0.5",
-        "@dhis2/ui-core": "^4.6.1",
+        "@dhis2/ui": "^5.0.2",
```
HendrikThePendric, line 31: We had `ui-widgets` as a dependency, and `ui-core` as a peerDependency. `ui` is now a peerDependency. Not quite sure if that is correct because I don't really know the rationale behind them being separated like that before.

PR #417, opened 2020-06-10
fix(init): let init produce app using css-modules instead of styled-jsx

HendrikThePendric:


cli/config/init.entrypoint.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import { DataQuery } from '@dhis2/app-runtime'
 import i18n from '@dhis2/d2-i18n'
+import classes from './init.App.module.css'
```
amcgee, line 4: This needs to be `./App.module.css`, it's renamed when the template is actually installed

cli/src/commands/init.js
```diff
@@ -197,6 +197,10 @@ const handler = async ({ force, name, cwd, lib }) => {
                 path.join(__dirname, '../../config/init.App.test.js'),
                 path.join(paths.base, 'src/App.test.js')
             )
+            fs.copyFileSync(
+                path.join(__dirname, '../../config/init.App.module.css'),
+                path.join(paths.base, 'src/App.module.css')
```
amcgee, line 202: See here - it's renamed on init.

cli/config/init.entrypoint.js
HendrikThePendric: Thanks for spotting that. I'm pretty sure I typed the file name without `init.`. I suspect my VSCode import plugin has tried to correct me 🤦  .... Anyhow, it's corrected now.

PR #421, opened 2020-06-18
fix(dependency): update babel-preset-env

### LINKED ISSUE #420: Babel preset env error with node 12+
I'm running into a build error because of a bug in babel-preset-env: https://travis-ci.com/github/dhis2/scheduler-app/builds/172025275#L277.

I traced it down to this issue https://github.com/nodejs/node/issues/32852. Which can be solved by:

> For those babel users who are affected by this issue after upgrading to node.js 13.13/12.17, please upgrade @babel/preset-env to 7.9.0.

https://github.com/nodejs/node/issues/32852#issuecomment-613655150

So to fix the above we'll need upgrade babel-preset-env. I see we have it listed in our dependencies directly for cli-app-scripts, and indirectly in app-shell. I assume just upgrading the version in cli-app-scripts will be sufficient (and the other willbe deduplicated to the newer version).

ismay:
Closes #420

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -29,7 +29,7 @@
         "@babel/plugin-syntax-dynamic-import": "^7.8.3",
         "@babel/plugin-syntax-nullish-coalescing-operator": "^7.8.3",
         "@babel/plugin-syntax-optional-chaining": "^7.8.3",
-        "@babel/preset-env": "^7.6.2",
+        "@babel/preset-env": "^7.9.0",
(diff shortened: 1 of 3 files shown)


amcgee:
@ismay thanks for this!  It looks like it's missing the `yarn.lock` update though...?

ismay:
> @ismay thanks for this! It looks like it's missing the `yarn.lock` update though...?

Ah yeah you're right, forgot about that. Updated the lockfile in e17f47f

PR #439, opened 2020-08-14
feat: Require @dhis2/ui v5

amcgee:
BREAKING CHANGE: Requires @dhis2/ui@^5


amcgee:
We could possibly require `@dhis2/ui@>5` instead of `^5` if we want to be lenient when a v6 is released, but we can make that a separate non-breaking change.  We also might want to move external deps to PeerDeps, though that eliminates the advantage of hoisting all standard deps from app-platform... will defer that to a later decision.

PR #448, opened 2020-08-17
chore: merge approved and passing dependabot prs

varl:
@amcgee we could try this action: https://github.com/ridedott/merge-me-action


.github/workflows/merge-me.yml
```diff
@@ -0,0 +1,29 @@
...
+    merge-me:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Merge me!
+              uses: ridedott/merge-me-action@master
+              with:
+                  # This must be used as GitHub Actions token does not support
+                  # pushing to protected branches.
+                  GITHUB_LOGIN: dependabot
+                  GITHUB_TOKEN: ${{secrets.GH_TOKEN}}
```
amcgee, line 28: We should probably use `secrets.DHIS2_BOT_GITHUB_TOKEN` (the org secret) here right?

.github/workflows/merge-me.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+    merge-me:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Merge me!
+              uses: ridedott/merge-me-action@master
+              with:
+                  # This must be used as GitHub Actions token does not support
+                  # pushing to protected branches.
+                  GITHUB_LOGIN: dependabot
```
amcgee, line 27: Is `dependabot` a valid user here?  Should it not be `dhis2-bot`?

.github/workflows/merge-me.yml
amcgee: (to be fair the other actions in this repo still use the local GH_TOKEN secret, so this is consistent and should be eventually changed for all the actions)

varl: Yeah, `DHIS2_BOT_GITHUB_TOKEN` is the right one to use.

.github/workflows/merge-me.yml
varl: The idea is to allow `dependabot` to authenticate with the token for `dhis2-bot` if I understand the workaround correctly. `dependabot` is the default value if omitted, but I figure that it is preferable to be explicit.

The `GITHUB_LOGIN` is used to determine which PRs from a user to merge, so if we want to automatically merge PRs by `dhis2-bot` we should put `dhis2-bot` as `GITHUB_LOGIN`. In this case we want to merge PRs by `dependabot`.

We may want to use a another action for auto-merging e.g. translations and/or other PRs from dhis2-bot in the future.

amcgee: Ah got it, that makes sense!

varl:
Looks like it works now:

```
Automatic merges enabled for GitHub login: dependabot.
Pull request created by varl, not dependabot, skipping.
```

varl:
I think Kodiak is a better option. Closing.

PR #451, opened 2020-09-01
feat: add d2-app-scripts deploy command

amcgee:
This adds the `d2-app-script deploy` command.  It takes an existing application bundle (created with `d2-app-scripts build`) and pushes it to a running DHIS2 instance.  For example:

Screen Shot 2020-09-01 at 7 05 08 PM


cli/src/commands/deploy.js
```diff
@@ -0,0 +1,186 @@
...
+            type: 'input',
+            name: 'baseUrl',
+            message: 'DHIS2 instance URL:',
+            when: () => !params.baseUrl,
+            validate: input =>
+                !input ||
+                !input.length ||
+                (!input.startsWith('http://') && !input.startsWith('https://'))
+                    ? 'Please enter a valid URL'
+                    : true,
```
HendrikThePendric, line 56: I'd find the reverse a lot easier to digest:
```javascript
            validate: input =>
                input &&
                input.length &&
                (input.startsWith('http://') || input.startsWith('https://'))
                    ? true
                    : 'Please enter a valid URL',
```

cli/src/commands/deploy.js
```diff
@@ -0,0 +1,186 @@
...
+}
+
+const handler = async ({ cwd = process.cwd(), ...params }) => {
+    const paths = makePaths(cwd)
+    const config = parseConfig(paths)
+
+    params = {
+        ...params,
+        ...(await promptForInputs(params)),
+    }
```
HendrikThePendric, line 80: Unless you are fairly certain that things will become more complex in the future, and more properties will be added to `params`, I'd like it if this would be a little bit more explicit:

```javascript
const { baseUrl, username, password } = {
    ...params,
    ...(await promptForInputs(params)),
}
```

cli/src/lib/httpClient.js
```diff
@@ -0,0 +1,10 @@
+const axios = require('axios').default
+
+module.exports.createClient = ({ baseUrl, auth, ...options }) => {
+    return axios.create({
+        baseURL: baseUrl,
+        timeout: 30000,
```
HendrikThePendric, line 6: I am blissfully unaware of the context, but a 30 second timeout seems pretty extreme. Is this because it just takes a long time for an app to get"installed" onto an instance?

HendrikThePendric (commented):
Looks great, I left a couple of review comments. I mostly focussed on code-style, because I haven't worked on these type cli-scripts myself, so I can't really provide feedback on the general approach. I'll leave it up to another, more knowledgeable, colleague to either approve or request changes.

cli/src/lib/httpClient.js
erikarenhill: I think this is as the client upload a zip-file to the instance, consider someone on a slow connection uploading a couple of megabytes?

amcgee: Yeah exactly.  I'll remove the timeout from the base client object and just specify it on the upload request.

cli/src/commands/deploy.js
```diff
@@ -0,0 +1,186 @@
...
+const path = require('path')
+const fs = require('fs-extra')
+const FormData = require('form-data')
+const inquirer = require('inquirer')
+
+const makePaths = require('../lib/paths')
+const parseConfig = require('../lib/parseConfig')
+const { createClient } = require('../lib/httpClient')
+
+const constructAppUrl = (baseUrl, config, serverVersion) =>
```
erikarenhill, line 11: would be nice with some comments/docs for these long nested ternary blocks of code to make it more understandable , or split up to a normal block, I find this quite hard to read now without really focusing and putting some time in it.

Might be enough with some jsdoc or 2-3 example outputs of the url depending on the version <35 or >= 35

amcgee: Updated (moved to a separate file, got rid of the complex ternary, added comments to clarify why behavior is different) let me know if it's easier to understand!

amcgee: https://github.com/dhis2/app-platform/pull/451/files#diff-0fac56a67ec46d47cec4c17ae9c1db6b

cli/src/commands/deploy.js
amcgee: Done, agreed

cli/src/commands/deploy.js
amcgee: Done, factored into helper function

amcgee:
@erikarenhill @HendrikThePendric thanks for the reviews, updated

HendrikThePendric (approved):

erikarenhill (approved):

PR #459, opened 2020-09-09
fix: base url should be overridden with env var if present

### LINKED ISSUE #458: Build command not picking up env variable override
I was expecting the build command to pick up on my override for the `DHIS2_BASE_URL` (https://github.com/dhis2/app-platform/blob/master/cli/src/commands/build.js#L40). It's been set correctly in my env, and is available for child processes, but the build command is not picking it up:

Screenshot 2020-09-09 at 10 56 46

ismay:
Closes #458 

I looked around, but this seems to be the only place where this pattern was used. I chose a separate var instead of using parentheses to clarify the intent, so that later on someone doesn't accidentally remove the parentheses.

### DIFF
--- a/cli/src/commands/build.js
+++ b/cli/src/commands/build.js
@@ -37,8 +37,9 @@ const setAppParameters = (standalone, config) => {
         standalone === false ||
         (typeof standalone === 'undefined' && !config.standalone)
     ) {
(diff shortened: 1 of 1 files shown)


amcgee (approved):
🎉  LGtM

PR #466, opened 2020-10-06
fix: adjust formatting in locales.hbs so it matches ESLint rules

HendrikThePendric:
This is primarily just a quick fix to prevent the `yarn lint` command from listing these formatting errors in app-platform apps.


cli/src/lib/i18n/templates/locales.hbs
```diff
@@ -1,20 +1,20 @@
...
 
-const namespace = '{{ namespace }}';
+const namespace = '{{ namespace }}'
 {{#each langs as |lang key|}}
-i18n.addResources('{{ lang }}', namespace, {{ lang }}Translations);
+i18n.addResources('{{ lang }}', namespace, {{ lang }}Translations)
 {{/each}}
 
-export default i18n;
+export default i18n
```
amcgee, line 20: Shouldn't this have a newline at the end?

amcgee:
Nice, thanks @HendrikThePendric !

I think this is fine and a good plan, however platform apps *shouldn't* break on files in `src/locales` since it's in `.gitignore`... at least I haven't seen it getting triggered except when that file is incorrectly in the `git` index... which app is triggering this eslint failure?

HendrikThePendric:
> however platform apps _shouldn't_ break on files in `src/locales`

Just to be clear: no nothing breaks, you're completely right. I'll explain what triggered me to create this PR:
When porting apps to the platform there is a step where you apply the new formatting (ESLint & Prettier) rules and inevitably end up fixing lots of ESLint errors. When working on those I find myself running `yarn lint` quite often, and it'd be nice if the list of js-errors wasn't cluttered by the errors in `src/locales/index.js`.

amcgee:
Right, but I think `lint` should ignore the files in `src/locales`...?

HendrikThePendric:
Yeah good point.... I hadn't really seen this happen in other apps I ported to the app-platform. I've just verified two things:
A. `src/locales` is present in the `.gitignore`, [see here](https://github.com/dhis2/messaging-app/blob/master/.gitignore#L19)
B. When I run `yarn lint` I do see these lint errors coming from `src/locales/index.js`

So not sure what's going there... However, regardless of all of this it would still be nice if the formatted file is compliant with our current set of lint rules.

amcgee:
@HendrikThePendric looks like that app [has a custom .eslintignore](https://github.com/dhis2/messaging-app/blob/master/.eslintignore) - maybe try removing it

amcgee:
See [cli-style docs on ignore files](https://cli-style.dhis2.nu/#/ignore-files)

amcgee (approved):
Change LGTM!

HendrikThePendric:
> @HendrikThePendric looks like that app [has a custom .eslintignore](https://github.com/dhis2/messaging-app/blob/master/.eslintignore) - maybe try removing it

Yes that was it 🎉  Thanks for that.

PR #467, opened 2020-10-07
fix: encode username and password when posting to server in AuthBoundary

awgaan:
Can now correctly log in with usernames/passwords that contain URL sensitive characters.


amcgee (approved):
Nice catch @awgaan !  LGTM

Issue #369, opened 2020-04-22
[spec] i18n revamp (WORK IN PROGRESS) 


amcgee:
**THIS IS A WORK IN PROGRESS**, I will continue fleshing it out before sharing for comment

I'm opening this issue to discuss an overhaul of the internationalization (**i18n**) system.  I'll start by laying out the challenges we face with the current system, and this issue will then coalesce into a specification for an overhaul of the platform-native i18n functionality.

#  The Problem(s)

## Types of translations in DHIS2

There are three primary categories of strings in 
1. **Server-bundled translations** For a long time, all translations were bundled into the DHIS2 core and accessible through an API endpoint.  Pretty much all use of this system has been eliminated, but some strings still have their translations baked into the WAR file (notably bundled application names)

2. **Application-bundled translations** 

3. **Dynamic translations** Some strings are truly dynamic, created by the implementer of the instance when configuring their metadata model.  These strings are created, updated, and accessed through the API and stored in the database.  We won't be directly addressing this use-case in for application-bundled strings, but it's good to keep in mind.

## A brief history of application-bundled translations

1. **d2 (the ancient solution)** -

2. **d2-i18n (the legacy solution)** - 

3. **d2-app-scripts i18n (the current solution)** - 

There are a few straggling uses of the (relatively) ancient `d2` system in some of our libraries (notably `d2-ui`).  I won't address the challenges with 

## Challenges with the `d2-i18n` "legacy" solution

Until very recently, the "standard" fleet of internationalization tools for DHIS2 libraries and applications was :

- [d2-i18n](https://github.com/dhis2/d2-i18n)
- [d2-i18n-extract](https://github.com/dhis2/d2-i18n-extract)
- [d2-i18n-generate](https://github.com/dhis2/d2-i18n-generate)

...TODO

## Challenges with the "current" `d2-app-scripts i18n` system

This is almost a direct port of the previous system.  It combines ported versions of `d2-i18n-extract` and `d2-i18n-generate` into a single App Platform cli module (`d2-app-scripts i18n`) and automatically runs the extract and generate steps when building an application or library with the platform.  The components are:

- [d2-i18n](https://github.com/dhis2/d2-i18n)
- [d2-app-scripts i18n extract](https://github.com/dhis2/app-platform/tree/master/cli/src/lib/i18n/extract.js)
- [d2-app-scripts i18n generate](https://github.com/dhis2/app-platform/tree/master/cli/src/lib/i18n/generate.js)

...TODO

[...]

PR #469, opened 2020-10-07
revert: keep default values that may have been set (#382)

jenniferarnesen:
This reverts commit 4b6e275969713014a9958cef0be256a00f16ef9a.

Upload to transifex fails when there are translations in the pot file, which is what this commit introduced. We'll have to solve translations for plurals in a different way.


amcgee:
Should we instead generate an `en.po` file?  Would that work with Transifex if `en.pot` also exists?

jenniferarnesen:
> Should we instead generate an `en.po` file? Would that work with Transifex if `en.pot` also exists?

@amcgee I added a comment to this issue explaining what I've found. https://github.com/dhis2/app-platform/issues/369#issuecomment-704839627 Would very much appreciate input.

jenniferarnesen:
> Should we instead generate an `en.po` file? Would that work with Transifex if `en.pot` also exists?

Good suggestion. We would probably want to get rid of the pot file then.

amcgee:
> Good suggestion. We would probably want to get rid of the pot file then.

Don't we need this for Transifex?

amcgee (approved):
LGTM, though we'll want to figure out a better solution...

PR #470, opened 2020-10-09
feat: detect server version, use latest available api version

amcgee:
This change deprecates the `DHIS2_API_VERSION` environment variable, which would previously hard-code the api version to be used by every runtime query in the application.  The previous default was 32, which is several versions behind at this point.

After this change, the App Shell makes a single call to an un-versioned API endpoint `/api/system/info` to detect the version of the DHIS2 instance (as well as some other immutable system information).  The server version is then parsed to determine the "latest" version of the DHIS2 API to use - equal to the minor server version (this may need to change in the future, if we slow down breaking API changes).

Additionally, the results of the `/system/info` all are saved into the `systemInfo` property of the App Runtime config service (https://github.com/dhis2/app-runtime/pull/652), and the parsed server version (`major`, `minor`, `patch`, `tag`) are saved in the `serverVersion` property.

Finally, this cleans up and refactors the folder hierarchy of the `app-adapter` package.

I added some test and noticed that neither build nor test were previously run on CI, so I updated the GHA workflow to run lint/build/test in parallel in a single workflow.


varl (approved):
Very nice feature addition! Also great to get rid of the hardcoded fallback to 32. :+1:

PR #471, opened 2020-10-19
feat: generate d2.config.json from d2.config.js

amcgee:
This generates a new config file `d2.config.json` from the compiled [application config](https://platform.dhis2.nu/#/config/d2-config-js-reference) (merge of `package.json` and `d2.config.js`).  The `.json` file is placed at the root of the application output, next to `manifest.webapp`, which it will eventually replace.  The data in this config file is necessary to enforce immutable application slugs, parse version numbers, and other things when publishing to the App Hub.

Here is a sample `d2.config.json` output generated from `examples/simple-app` :

```json
{
  "name": "simple-app",
  "title": "Simple Example App",
  "description": "This is a simple example application",
  "standalone": true,
  "type": "app",
  "version": "1.0.0",
  "author": {
    "name": "Austin McGee",
    "email": "austin@dhis2.org"
  }
}
```

### DIFF
--- a/cli/src/lib/generateManifest.js
+++ b/cli/src/lib/generateManifest.js
@@ -32,5 +32,12 @@ module.exports = (paths, config, publicUrl) => {
 
     reporter.debug('Generated manifest', manifest)
 
+    // For backwards compatibility, WILL BE DEPRECATED
     fs.writeJsonSync(paths.buildAppManifest, manifest, { spaces: 2 })
+
+    // Write config json
+    const appConfig = { ...config }
+    delete appConfig['entryPoints']
+
+    fs.writeJsonSync(paths.buildAppConfigJson, appConfig, { spaces: 2 })
 }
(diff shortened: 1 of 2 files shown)


varl (approved):
Short and sweet.

Question: will it be more common to include or omit properties in the json output?

amcgee:
> Question: will it be more common to include or omit properties in the json output?

I think it should be rare to do either.... the `endPoints` property is omitted here because it references an entrypoint in source which doesn't exist in the compiled output, most everything else is configuration options which serve as input to the compiler and metadata to the compiled app

amcgee:
@varl does that seem reasonable?  Should we explicitly construct the `d2.config.json` here to more clearly define its structure?

amcgee:
Some additions we might consider... `icon`, `buildTimestamp`, ...

amcgee:
Some exclusions we might consider... `type`, `standalone`

varl:
I think it's reasonable to just delete the key now, and when we need a more sophisticated schema for the json, we can create a function that constructs the `d2.config.json` and handle inclusions/exclusions there.

PR #472, opened 2020-10-21
feat: Initialize locales and generate translations in the shell

amcgee:
BREAKING CHANGE: Applications must delete the `/src/locales` directory and remove all imports referencing `src/locales`

Generate translations and load locale in the App Shell rather than in the application.  Previously, the `/src/locales` directory in application source was automatically generated by the platform compiler and needed to be manually imported from application code.  The `/src/locales` directory was also typically included in `.gitignore`, causing confusion since a non-existent file might be referenced from App.js.

**TODO**
- [ ] ensure functionality with library builds, including library localization rollup
- [ ] dynamically load user locale strings in parallel with app code, rather than loading everything (big bundle size win!)
- [ ] (maybe) add support for translations in non-e2e tests


varl (approved):

varl:
Added approval for this entire thing, expecting to re-review when it's "ready for review".

PR #475, opened 2020-10-30
fix: update react-scripts and cli-style at same time to use eslint 7

varl:
Update to newest cli-style and react-scripts to ensure that we get the improvements made in react-scripts for the changes in eslint related to resolving plugins.

Supersedes #473


amcgee:
@varl looks good, but won't this circumvent the "we manually extend the existing ESLint configuration from the project in the shell ESLint configuration", since the shell will no longer use the `eslintrc.js` config?

varl:
You're right, it does. I was under the impression that this would allow ESLint to load the `.d2/shell/.eslintrc.js`. It doesn't since `EXTEND_ESLINT` is what makes `react-scripts` allow extensions to the built-in ESLint configuration.

amcgee:
OK yeah.  Probably if we’re doing that we should delete the eslint config too?  It’s nice to have when developing the shell, but probably not critical

varl:
Yeah. I'll see if I can dig a bit more.

varl:
Actually, react-scripts @ 4 might work for us:

varl:
@amcgee I'd like to update cli-style and react-scripts to the latest versions in app-platform, and release it on `alpha` to test it out fully. It does the trick locally.

package.json
```diff
@@ -8,27 +8,29 @@
     },
     "workspaces": {
         "packages": [
+            "examples/simple-app",
```
amcgee, line 11: Why did we decide to do this in this change?  I think there were some issues I tried it previously... don't remember exactly what they were

package.json
```diff
@@ -8,27 +8,29 @@
...
         "format": "d2-style js apply && d2-style text apply",
         "lint": "d2-style js check && d2-style text check",
-        "build:adapter": "cd adapter && yarn build",
-        "build": "yarn build:adapter && yarn install:example",
-        "install:example": "cd examples/simple-app && yarn --force --check-files",
-        "start:example": "cd examples/simple-app && yarn start --force",
+        "build": "yarn build:adapter && yarn build:example",
+        "build:adapter": "yarn workspace @dhis2/app-adapter build",
+        "build:example": "yarn workspace simple-app build",
+        "install:example": "yarn workspace simple-app install  --force --check-files",
```
amcgee, line 28: If we're using the workspace here I don't think the install step is necessary

package.json
varl: I can revert it in `alpha` before we cut the proper release. The reason I went ahead with it was that I was having issues with the simple-app when I was testing out the shell changes, and this made it work.

PR #486, opened 2020-11-16
feat: introduce alerts component to show alerts as alert-bars in a stack

HendrikThePendric:
This is the component that goes with the alerts-service. It simply displays the alerts from the alert context as `AlertBars` in an `AlertStack`. The `onHidden` callback on the `AlertBar` will execute the `onHidden` passed as an option to `useAlert` and also the `remove` function on the `alert`, which will remove the `alert` from the context. `onHidden` is called from the `AlertBar` **after** the hide-animation completes, so things are hiding smoothly.

I'll leave some small notes in the code as well.

### DIFF
--- a/adapter/src/components/AuthBoundary.js
+++ b/adapter/src/components/AuthBoundary.js
@@ -28,6 +28,6 @@ export const AuthBoundary = ({ url, children }) => {
 }
 
 AuthBoundary.propTypes = {
-    children: PropTypes.element,
+    children: PropTypes.node,
     url: PropTypes.string,
 }
--- a/adapter/src/components/AuthBoundary.js
+++ b/adapter/src/components/AuthBoundary.js
@@ -28,6 +28,6 @@ export const AuthBoundary = ({ url, children }) => {
 }
 
 AuthBoundary.propTypes = {
-    children: PropTypes.node,
+    children: PropTypes.element,
     url: PropTypes.string,
 }
--- a/adapter/src/index.js
+++ b/adapter/src/index.js
@@ -16,8 +16,8 @@ const App = ({ url, apiVersion, appName, children }) => (
                 <HeaderBar appName={appName} />
                 <AuthBoundary url={url}>
                     <div className="app-shell-app">{children}</div>
-                    <Alerts />
                 </AuthBoundary>
+                <Alerts />
             </div>
         </ServerVersionProvider>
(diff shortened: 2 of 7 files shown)


adapter/src/index.js
```diff
@@ -15,6 +16,7 @@ const App = ({ url, apiVersion, appName, children }) => (
                 <HeaderBar appName={appName} />
                 <AuthBoundary url={url}>
                     <div className="app-shell-app">{children}</div>
+                    <Alerts />
```
HendrikThePendric, line 19: I felt that inserting `Alerts` here, as a child of the `AuthBoundary` and sibling of  `div.app-shell-app` was most appropriate. But that did force me to change prop-types on the `AuthBoundary`.

shell/package.json
```diff
@@ -13,9 +13,9 @@
     },
     "dependencies": {
         "@dhis2/app-adapter": "5.4.2",
-        "@dhis2/app-runtime": "^2.2.2",
+        "@dhis2/app-runtime": "^2.6.1",
         "@dhis2/d2-i18n": "^1.0.5",
-        "@dhis2/ui": "^5.4.2",
+        "@dhis2/ui": "^5.7.2",
```
HendrikThePendric, line 18: We needed a higher version of `@dhis2/ui` which includes https://github.com/dhis2/ui/pull/300 to avoid an infinite loop when showing more than one alert.

adapter/src/components/AuthBoundary.js
```diff
@@ -28,6 +28,6 @@ export const AuthBoundary = ({ url, children }) => {
 }
 
 AuthBoundary.propTypes = {
-    children: PropTypes.element,
+    children: PropTypes.node,
```
HendrikThePendric, line 31: See https://github.com/dhis2/app-platform/pull/486#discussion_r524033225

amcgee:
> I felt that inserting `Alerts` here, as a child of the `AuthBoundary` and sibling of  `div.app-shell-app` was most appropriate. But that did force me to change prop-types on the `AuthBoundary`. 

Hmmm... should we put it outside the Auth boundary?  That way we can show alerts when, i.e., an authentication attempt fails.

amcgee:
Looks great!  I think my only recommendation wouuld be to move the AlertStack outside the AuthBoundary

PR #491, opened 2020-12-04
feat: upgrade to yarn2 and improve compilation step

amcgee:
❓ Should we make this a breaking change?  I think maybe not

> **NB**: We could make this a non-breaking change by _including a copy of yarn2 in the shell itself, so that the install functionality there is completely self-contained_.  This might be a nice safety-hatch, but does come at a ~2MB shell bootstrapping cost.  I think even if we avoid this concrete breaking change we should bump the major version here, as there are a lot of improvements that _might_ have problems with yarn1 apps, particularly if those apps don't correctly declare their own dependencies (which many probably don't).

**EDIT** - I made this change - the shell is now completely self-contained and doesn't rely on the yarn version of the app under build.  A number of other things were also cleaned up, so it is tested and works with yarn 1 and yarn 2 apps.  A warning is shown when the app is running yarn1, but it is allowed.  The app scripts build and start now also validate the dependencies of the app's package.json to prevent duplicate singletons, and offers to move incorrect `dependencies` to both `peerDependencies` and `devDependencies`

**Biggest change**: upgrade to yarn v2 with plug'n'play (no more node_modules!) - [here's why](https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern)

> I originally enabled yarn's ["zero-install" feature](https://yarnpkg.com/features/zero-installs) experimentally by checking in all yarn cache zip archives (~75MB).  You would still need to run `yarn install` to build platform-specific packages, but it should be lightning fast.  We may not want to check in the cache as it does have some clone-size tradeoffs (and adds noise to PRs), but reproducibility should also quite improved since there cannot be dependency mismatches without git changes.  I've dropped that change from this PR though as it adds about 2000 files and drowns out the actual important changes, we can set it up later if we want.

This upgrade also checks in the `.yarn/releases/yarn-berry.cjs` file and references it in `.yarnrc.yml`, which will be picked up by the global `yarn@1` executable - no need to update anything on the developer system, it should **just work**

Other improvements made here:
- With the new `portal:` dependency protocol in yarn2 we [properly link the local app](https://github.com/dhis2/app-platform/pull/491/files#diff-3586ceb3c19adf7be9804c74e59b7c33b6c0082e739a8a2670b6c8c53e16ea4bR107-R111) **as a dependency of the local shell** while "following dependencies" (which basically means all dependencies will be installed and resolved / deduped in the shell, rather than in the app with the shell awkwardly as a depdency).  This is a huge win for explicitness in app dependencies (shell deps can be declared as peer deps or as explicit runtime deps, depending on need, and the shell will resolve them correctly) and should eliminate many "multiple copies of React" errors and other subtle duplication bugs.  This also supports stepping through application source files in browser debuggers (rather than the transpiled output) which is super nice.
- `@dhis2/app-shell` (`./shell` in the monorepo) is no longer included as a dependency of `@dhis2/cli-app-scripts` (`./cli`).  Instead, a copy of the shell is bundled with the `cli-app-scripts` package.  This is now possible because we run `yarn install` on a self-contained version of the shell when building with the platform, so we don't need to have all the transient dependencies bubble up (and we probably don't want them to anyway, as it is very error prone).  Another major advantage of this change is that we no longer have to install all the `app-shell` and `react-scripts` dependencies when installing the cli tool (and particularly the global `@dhis2/cli` which never needs them anyways)
- `d2-app-scripts build` and `d2-app-scripts start` now both accept `--shell` and `--adapter` parameters to support manual override of either the shell or the adapter used in a given app.  This should be an escape-hatch and isn't recommended (nor documented) but can be quite useful when waiting for a shell or adapter change to get published, for instance.  This is [used in the example app](https://github.com/dhis2/app-platform/pull/491/files#diff-ef40f2858134288d8f15bda09fdeb484295ee7ce8081d742e7be4f2875273ddcR23) to ensure it always uses the local versions of shell and adapter in development.
  - The `--shell` command-line argument accepts either a path to a local folder *or the name of a package* (which must be installed locally).  This would allow, for instance, using the latest `cli-app-scripts` with an older version of the shell (or vice versa) or even shimming in another published shell.  
  - The `--adapter` command-line argument accepts a `yarn` package source identifier.  Under the hood, the platform runs `yarn add @dhis2/app-adapter@<custom-adapter>` in the local shell directory, so it does not need to be installed in the local app.  Some cool supported options are: `portal:../../my-adapter` (to reference a local copy of `app-adapter`, with followed dependencies and auto-refresh on changes), `5.3.2` (to pin to a specific version), `npm:my-custom-adapter@^7` (to override with a different npm package), `dhis2/app-platform#head=my-feature-branch&workspace=@dhis2/app-adapter` (to pull the adapter from a feature branch of the dhis2/app-platform github)
- [Uses the new `patch:` protocol](https://github.com/dhis2/app-platform/pull/491/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R18) in yarn2 to apply a [patch file](https://github.com/dhis2/app-platform/pull/491/files#diff-8a2a9f25f3e8a4005b274a026074f8e76d60dcda3f88a13677ad0eae20fd1d0a) to the published `@dhis2/cli-style` (dhis2/cli-style#320) so that it works with yarn2
- Also apply a small [patch](https://github.com/dhis2/app-platform/pull/491/files#diff-f5a031e738442f9aba200c7248bdb4d93db5558b9edb1ac631233a6555ba8b20) to the published `react-dev-utils` (PR coming soon) so that it works with deep imports from dependent packages - this is configured as a deep-dependency [resolution in the shell's package.json](https://github.com/dhis2/app-platform/pull/491/files#diff-f3109cd7f20913370e87fa8c95643e0f3cb7f6aa6924caaee3bd5ab8d7ed498aR46)
- Specifies [a custom chunk name](https://github.com/dhis2/app-platform/pull/491/files#diff-f3817ea85bf647e55f82fac2ede7fa2a4af1cdc970be4845e3707fc23612ac0a) (`moment-<locale>`) when using dynamic imports for each `moment` locale, so that the chunk files are easily recognized on disk and in source explorer (previously they were numerically indexed next to other chunks and hard to identify)
- Uses the `packageExtensions` option in `.yarnrc.yml` to "patch" missing dependencies in `@dhis2/ui-*` packages (will open PRs shortly) since yarn2 is strict about declaring all used dependencies, and will bail out if undeclared dependencies are used.

TODO (possibly in follow-on PRs):
- [x] fix husky so that we can use hooks again - currently hooks break.  We should consider using the new husky v5 which supports an explicit `install` command (rather than installing hooks on post-install) and seems to support yarn v2
- [ ] `.eslintrc.js` in an app seems to break things if using a config other than `'react-app'`, which is strange (don't know why CRA is running eslint on portal dependencies...).  Should look into fixing this
- ~[ ] Need to submit the `react-dev-utils` patch to upstream `create-react-app` repo~ (soft links are impossible to separate from relative imports by the time they hit the webpack plugin, so we're just short-circuiting that plugin for now.  Definitely not an ideal solution.)
- [ ] Should fix dependency issues in our own upstream packages (`@dhis2/ui`, `@dhis2/cli-style`, `@dhis2/cli-utils` etc.)
- [ ] `@dhis2/cli-utils release` is broken (as are most of our other cli tools) because they don't resolve from the correct packages in pnp-land, need to fix this
- [x] Fix tree-shaking -- I think rollup is causing issues, we should probably just transpile module-by-module for libraries (See #492)
- [x] Enforce strict de-duplication of singleton dependencies in the shell/app at build time, and/or render the app as an independent React tree and "delegate" singleton references to the shell
- [ ] Consider enabling "zero-install" by checking in the yarn cache (deferred for now)
- [x] Consider downloading an isolated version of `yarn2` when bootstrapping the shell, instead of assuming that the version used by the app is also `>= 2.0`
- [ ] Consider running `@yarnpkg/doctor` during package verification in apps to ensure proper dependency declaration - portal deps appear not to always enforce these well
- [x] Fix sourcemaps to support debugging of source files (rather than compiled output) in Chrome and VSCode


amcgee:
To avoid requiring the app developer to use `yarn2` we could potentially check `yarn --version` when bootstrapping the shell, and if it's not v2 we could run `yarn set version latest` in the shell directory, which would set up an isolated yarn version for the shell only.  This still requires correct dependencies in the app's package.json, but wouldn't break if the app doesn't have yarn2 (since we rely on i.e. `portal:` dependencies in the shell)

amcgee:
Keeping this open as it's interesting to explore in the future

PR #492, opened 2020-12-07
feat: build unbundled libraries, validate package when building (LIBS-125)

amcgee:
This removes the diverging logic between application and library builds.  It no longer uses rollup, shares a single babel config, and compiles library files individually rather than bundling them into a single `lib.js` file (which is better for tree shaking)

This also adds the concept of package validation to `d2-app-scripts`.  This allows the build tool to show hints, offer automatic fixes, and bail on CI if the app or library being built is misconfigured.  Additional validations for singleton dependencies, yarn versions, and more could be added in the future (as has been done in #491 [here](https://github.com/dhis2/app-platform/pull/491/files#diff-66ced5f19aae04a3195731e4f70aca6529f61e4a1bbf86c6d7f261857456b3ea))

The first validator added here ensures that `package.main`, `package.module` and `package.exports` are set correctly for library builds - since the `lib.js` file is no longer generated, this will prevent an upgrade of `@dhis2/cli-app-scripts` from silently breaking library builds (which should mean we don't need to cut a breaking change here, though we could if we want to)

Example output with prompt to automatically fix the package export fields (declined here):

Screen Shot 2021-02-19 at 15 07 43

And another when `./build/<type>/lib.js` is incorrectly specified (prompt accepted here):

Screen Shot 2021-02-19 at 15 09 23


cli/config/makeBabelConfig.js
```diff
@@ -0,0 +1,74 @@
...
+        plugins: [
+            // Adds syntax support for import()
+            require('@babel/plugin-syntax-dynamic-import').default,
+
+            /*
+             * These plugins actually transform code
+             */
+
+            // Automatically include a React import when JSX is present
+            require('babel-plugin-react-require'),
```
mediremi, line 44: We might not need this thanks to the new JSX transform https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

amcgee: True!  However since we haven't bumped the required version of `react` yet (still `react@^16.8`) I think we need to keep it for now

cli/src/lib/compiler/compile.js
```diff
@@ -1,17 +1,140 @@
...
+    if (isApp) {
+        fs.removeSync(paths.shellPublic)
+        fs.copySync(paths.shellSourcePublic, paths.shellPublic)
+    }
+
+    const copyFile = async (source, destination) => {
+        await fs.copy(source, destination)
     }
+    const compileFile = async (source, destination) => {
+        if (path.extname(source) === '.js') {
```
mediremi, line 112: Should we also check if the `extname` is `.jsx` or `.ts`?

amcgee: Yep, nice catch!  Fixed

mediremi (approved):
LGTM :tada: 

Improvements in tree shaking are always nice and I really like the package validation + automatic fixes feature - which will be even more useful once we integrate validations for singleton dependencies from #491.

amcgee:
I don't think we need a breaking change here but @varl might disagree - will hold off on merging for a short bit to get more eyes on this.  We could also merge to `next` to start testing this out (will do that later today if no more comments come in)

amcgee:
I'm going to add some documentation then go ahead and merge this, as we want to get it rolling through the ecosystem.  If we find the "pseudo-break" too cumbersome we can back it out and cut a breaking change instead.

varl (commented):
Nice!

PR #496, opened 2021-01-11
chore: update cli-style and reorder app-shell imports

mediremi:
[`cli-style@7.3.0`](https://github.com/dhis2/cli-style/blob/master/CHANGELOG.md#730-2020-12-16) adds the eslint `import/order` rule, which ensures that imports are sorted alphabetically.

However, as the app-shell's files do not have their imports sorted, this will break `d2-app-scripts build` for any app that depends on `cli-style` version 7.3.0 (as shown by the screenshot below).

image


.eslintignore
```diff
@@ -2,4 +2,5 @@
 **/locales/index.js
 cli/config/init.entrypoint.js
 cli/config/init.App.test.js
-
+shell/src/D2App/*
+examples
```
amcgee, line 6: Why are we excluding `examples` from being linted?

.eslintignore
```diff
@@ -2,4 +2,5 @@
 **/locales/index.js
 cli/config/init.entrypoint.js
 cli/config/init.App.test.js
-
+shell/src/D2App/*
```
amcgee, line 5: The files in this directory should be valid, they are "placeholders" not generated by the compiler (they're replaced later)

amcgee (commented):
Added a couple questions - as I mentioned this morning, our eslint configurations in the shell and in apps using the platform should be isolated, so we shouldn't require AppShell code to conform to whatever eslint config the parent has (it could be anything, we don't enforce that it will always be derived from `cli-style`).

I'd prefer fixing that isolation, but it doesn't hurt to conform to our latest cli-style rules within this repo itself (when developing `app-platform`, not when using it in another repository)

.eslintignore
mediremi: I had previously encountered linting errors regarding unresolved imports due to './locales' not existing relative to `examples/simple-app/src/App.js` and did not want to include a `// eslint-disable-line import/unresolved` comment in an example. https://github.com/dhis2/app-platform/blob/9aea659cde05400b96bc0afca18c55ca6c0e886e/examples/simple-app/src/App.js#L2
This issue doesn't seem to appear anymore though when I remove `examples` from `.eslintignore` :man_shrugging:

amcgee (approved):
Looks good to me!  I'm not sure why the Semantic Pull Request never ran here, might have been a service outage...

amcgee:
I've updated this to be a `chore` which won't cut a new patch release.  As I mentioned above we can't make assumptions about the eslint environment of the calling app, so this change is just about updating the `app-platform` development environment.  I'll test this in a new platform app and if needed fix the eslint isolation to ensure this works (I believe #491 does this already, but in a different way as it enables full app-shell isolation)

amcgee:
Nice work @mediremi

PR #502, opened 2021-02-18
fix: pin styled-jsx@3.3.2 to avoid styled-jsx/babel bug

amcgee:
There is an open [bug in styled-jsx/babel](https://github.com/vercel/styled-jsx/issues/695) introduced in 3.3.3 which causes invalid js output.  Until that issue is fixed, pin our version of styled-jsx to 3.3.2 (for compilation, runtime can still track later versions)

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -61,7 +61,7 @@
         "rollup-plugin-postcss": "^2.0.3",
         "rollup-plugin-replace": "^2.2.0",
         "rollup-plugin-visualizer": "^3.2.2",
-        "styled-jsx": "^3.2.5"
+        "styled-jsx": "<3.3.3"
     },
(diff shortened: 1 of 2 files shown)


jenniferarnesen (approved):

jenniferarnesen:
Very minor - should the description say pin it to 3.3.2, or pin it to under 3.3.3 (instead of 3.3.3)?

martinkrulltott (approved):

martinkrulltott:
> Very minor - should the description say pin it to 3.3.2, or pin it to under 3.3.3 (instead of 3.3.3)?

I agree, changed it to:

> Until that issue is fixed, pin our version of styled-jsx to 3.3.**2**

PR #517, opened 2021-02-25
fix: convert jsx, ts, and tsx files to js on build

amcgee:
Follow-on to #492 fixing support for typescript source files (app-runtime)


mediremi (approved):

cli/src/lib/validators/validatePackageExports.js
```diff
@@ -6,6 +6,7 @@ const { writeJSON } = require('fs-extra')
  * Ensure that package.main, package.module, and package.exports are valid
  */
 
+const normalizeExtension = ext => ext.replace(/\.[tj]sx?$/, '.js')
```
Mohammer5, line 9: Not necessary, but could be used in the above file as well

Mohammer5 (approved):

cli/src/lib/validators/validatePackageExports.js
amcgee: Yeah, wasn't sure it was worth it but is nicer with shared code - moved to helper module in b0dec7a

Mohammer5 (approved):

PR #519, opened 2021-02-26
fix: allow for nested entry-points under src

varl:
This change makes it so that we resolve the entry-point relatively to
the src folder, so src/react/index.js becomes react/index.js.

This resolves an issue when the library entry-point is not defined as
src/index.js.

So for example src/react/index.js will fail to build because we only
used the basename (index.js) to construct the build entry-point path, so
the react part of the path disappears from the expected path.

We got build/es/index.js when we should have gotten
build/es/react/index.js.

Since we force users of app-scripts to put their source code under
'src' (compiler/compile.js:15), we can safely assume that the configured
entry-point can be resolved relatively to the src directory and grab
everything to the right and build the path to the target directory, e.g.
build/es/react/index.js.

### DIFF
--- a/cli/src/lib/validators/validatePackageExports.js
+++ b/cli/src/lib/validators/validatePackageExports.js
@@ -31,21 +31,20 @@ module.exports.validatePackageExports = async (
     const baseDir = path.dirname(paths.package)
 
     let valid = true
-    const entrypointBasename = normalizeExtension(
-        path.basename(config.entryPoints.lib)
-    )
+    const entrypointBasename = normalizeExtension(config.entryPoints.lib)
+    const relativeEntrypoint = path.relative('src', entrypointBasename)
 
     const expectedESMExport =
         './' +
         path.relative(
             baseDir,
-            path.join(paths.buildOutput, 'es', entrypointBasename)
+            path.join(paths.buildOutput, 'es', relativeEntrypoint)
         )
     const expectedCJSExport =
         './' +
         path.relative(
             baseDir,
-            path.join(paths.buildOutput, 'cjs', entrypointBasename)
+            path.join(paths.buildOutput, 'cjs', relativeEntrypoint)
         )
 
     const expectedPackage = {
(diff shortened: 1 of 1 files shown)


Mohammer5 (approved):

cli/src/lib/validators/validatePackageExports.js
```diff
@@ -31,21 +31,20 @@ module.exports.validatePackageExports = async (
     const baseDir = path.dirname(paths.package)
 
     let valid = true
-    const entrypointBasename = normalizeExtension(
-        path.basename(config.entryPoints.lib)
-    )
+    const entrypointBasename = normalizeExtension(config.entryPoints.lib)
+    const relativeEntrypoint = path.relative('src', entrypointBasename)
```
amcgee, line 35: Should use `paths.src` instead of hard-coding `'src'`

amcgee (approved):
Minor change suggested, after that looks good.  Thanks for fixing this @varl and for finding it @Mohammer5!

cli/src/lib/validators/validatePackageExports.js
```diff
@@ -31,21 +31,20 @@ module.exports.validatePackageExports = async (
     const baseDir = path.dirname(paths.package)
 
     let valid = true
-    const entrypointBasename = normalizeExtension(
-        path.basename(config.entryPoints.lib)
-    )
+    const entrypointBasename = normalizeExtension(config.entryPoints.lib)
```
amcgee, line 34: Also - this variable is no longer the `basename` so it should be renamed (just `entrypoint` should be fine)

cli/src/lib/validators/validatePackageExports.js
amcgee:
```suggestion
    const relativeEntrypoint = path.relative(paths.src, entrypointBasename)
```

cli/src/lib/validators/validatePackageExports.js
varl: Dropped the interim variable in: 2dda1d9

cli/src/lib/validators/validatePackageExports.js
varl: Thanks! 2dda1d9]

varl:
Thanks for the reviews. I'll merge when it passes.

PR #521, opened 2021-02-26
fix: allow jest to ignore the build folder

varl:
Now that we only use Babel to transpile sources to CJS and ES, we also
copy over everything in the source directory to the build directly.

Using UI as an example, that includes for example:

- `**/features/*`
- `**/*.test.js`
- `**/*.stories.js`
- `**/*.stories.e2e.js`

The test files proved problematic because Jest picks them up from the
build folder as well as the source folder, causing the tests to be run
thrice: src, build/es, and build/cjs.

Furthermore, it is not necessary true that tests can be transpiled and
still work as expected. We saw this in UI as well, where the test in src
passes, but the test in build/cjs fails.

As a first step this fixes the test command so that Jest ignores tests
under the build folder.

### DIFF
--- a/cli/config/jest.config.js
+++ b/cli/config/jest.config.js
@@ -3,6 +3,7 @@ module.exports = {
     transform: {
         '^.+\\.[t|j]sx?$': require.resolve('./jest.transform.js'),
     },
+    modulePathIgnorePatterns: ['build'],
     moduleNameMapper: {
         '\\.(css|less)$': require.resolve('./jest.identity.mock.js'),
         '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$': require.resolve(


varl:
The second step is to not copy over all the unnecessary files to the `build`. I started with that implementation and allowing Babel to ignore certain files (pretty much the list in the description), but that causes `babel-jest` to become annoyed that Babel is ignoring certain files, but Jest is not, so it requires us to add the `transformIgnorePatterns` to the Jest configuration.

I've had experience with that being a very fiddly option to get right, so I determined that simply ignoring the `build` folders in Jest's test finder is a simpler and safer fix for now.

Mohammer5 (approved):

amcgee:
I don't think this is the right change.  We specify `./src` as the root for jest in the [test command](https://github.com/dhis2/app-platform/blob/master/cli/src/commands/test.js#L33), so in normal applications this change would result in ignoring `./src/build` which is incorrect.  The issue in `ui` is that it is a monorepo and the `d2-app-scripts test` command is run from the monorepo root.  The `jest` command specifies `./packages` as the root in [its `jest.config.js` override](https://github.com/dhis2/ui/blob/master/jest.config.js#L4), which will include a ton of unnecessary files.  The correct fix in the `ui` case for jest is to specify `./packages/*/src` as the root instead, or alternatively run the test command in each individual package.

amcgee:
Feel free to reopen if you disagree

PR #525, opened 2021-03-01
fix: don't render application until we are sure the locale is initialized

amcgee:
Because we set the locale state in a `useEffect`, there was a subtle race condition bug which could cause the application to render before the locale is set.  When this happens, strings in the app are not correctly translated because the locale hasn't yet been switch from English to whatever the user has selected.

### DIFF
--- a/adapter/src/components/AuthBoundary.js
+++ b/adapter/src/components/AuthBoundary.js
@@ -13,9 +13,11 @@ const settingsQuery = {
 
 export const AuthBoundary = ({ url, children }) => {
     const { loading, error, data } = useDataQuery(settingsQuery)
-    useLocale(data && data.userSettings.keyUiLocale)
+    const locale = useLocale(
+        data && (data.userSettings.keyUiLocale || window.navigator.language)
+    )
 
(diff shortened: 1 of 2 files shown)


mediremi (approved):

adapter/src/components/AuthBoundary.js
```diff
@@ -13,9 +13,11 @@ const settingsQuery = {
 
 export const AuthBoundary = ({ url, children }) => {
     const { loading, error, data } = useDataQuery(settingsQuery)
-    useLocale(data && data.userSettings.keyUiLocale)
+    const locale = useLocale(
+        data && (data.userSettings.keyUiLocale || window.navigator.language)
```
HendrikThePendric, line 17: Is this wise?
I would have expected the DHIS2 user-locale to take precedence over the browser language.

HendrikThePendric: Did you perhaps mean to write:
```js
(data && data.userSettings.keyUiLocale) || window.navigator.language
```

varl (approved):

PR #530, opened 2021-03-09
feat: upgrade to @dhis2/ui v6 and bump other dependency versions

amcgee:
Related to [DHIS2-9893](https://jira.dhis2.org/browse/DHIS2-9893)

This bumps @dhis2/app-platform to use ui v6 which should have all the latest and greatest headerbar fixes.

BREAKING CHANGE: applications must use `@dhis2/ui@^6.5.3`

### DIFF
--- a/adapter/package.json
+++ b/adapter/package.json
@@ -37,7 +37,7 @@
     "peerDependencies": {
         "@dhis2/app-runtime": "^2",
         "@dhis2/d2-i18n": "^1",
-        "@dhis2/ui": "^5",
+        "@dhis2/ui": "^6",
         "classnames": "^2",
         "moment": "^2",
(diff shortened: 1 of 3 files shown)


varl:
1. Yes, getting dhis2/app-runtime#788 in first would be nice since it is approved and would be fantastic to get propagated to all platform apps.
2. Let's stick to releasing "technically" breaking changes as breaking changes, even if they do not require the user to change things. This discussion comes up every now and then, so I think we should stick to the process.

amcgee:
@varl since we've addressed the points above and this has your tacit approval I'm merging it to `next` to test propagation to apps.  You can approve or reject the subsequent merge to master.

PR #532, opened 2021-03-11
feat(publish): upload apps to App Hub

Birkbjo:
This adds a command that is pretty similar to `deploy`. It extracts some information from the built `appBundle`, however this uploads that bundle to App Hub instead of a DHIS2-instance. There's some code that is duplicated from `publish` (eg. `dumpHttpError`), could probably extract some of this to a common file. I didn't want to change more code than necessary.

This should mainly be used with app-platform apps, and it's thus very simple to use with a bundled app. However I added an escape hatch so that apps that are not migrated to the platform may use this to upload any arbitrary bundle to the app hub. That's why we have options like `appId`, `file` and `file-version`. There are not intended to be use for overriding the config from a built app, as we do not want discrepancies from the uploaded app-information and the app-manifest. 

### How to test it?
 API key generation is live on both staging and live server.

- Log in to https://staging.apps.dhis2.org/ 
- Navigate Api Key, and generate a new key
- Use API-key with `--apiKey` or set as env with `D2_apiKey=apiKey` when using the CLI.

TODO:

- [x] Initial implementation
- [x]  Initial docs

### DIFF
--- a/cli/src/commands/publish.js
+++ b/cli/src/commands/publish.js
@@ -42,8 +42,8 @@ const validateFields = (
     fieldNames.forEach(fieldName => {
         const fieldValue = config[fieldName]
         if (
-            (requiredFields.has(fieldName) && fieldValue == null) ||
-            fieldValue === ''
+            requiredFields.has(fieldName) &&
+            (fieldValue == null || fieldValue === '')
         ) {
             reporter.error(
                 `${fieldName} not found in config. Add an ${chalk.bold(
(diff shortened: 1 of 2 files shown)


cli/src/commands/publish.js
```diff
@@ -0,0 +1,216 @@
...
+            validate: v =>
+                isValidServerVersion(v) ? true : 'Invalid server version',
+        },
+        {
+            type: 'input',
+            name: 'maxVersion',
+            message: 'Maximum DHIS2 version supported',
+            when: () => !params.maxVersion,
+            validate: v =>
+                !v || isValidServerVersion(v) ? true : 'Invalid server version',
```
mediremi, line 112: Maybe the DHIS `min` and `max` versions should be specified in `d2.config.js` instead of as command parameters :thinking: It would make publishing through CI easier as the publish command and its parameters run by the CI action would not have to change

amcgee: 👍  agreed, I think that would be much cleaner.  We can keep this as-is for now and add that later if we want (@Birkbjo would you prefer to do it now or in a separate change?)

cli/src/commands/publish.js
```diff
@@ -0,0 +1,216 @@
...
+
+const promptForConfig = async params => {
+    if (process.env.CI && (!params.apikey || !params.minVersion)) {
+        reporter.error(
+            'Prompt disabled in CI mode - missing apikey or minVersion parameter.'
+        )
+        process.exit(1)
+    }
+
+    const responses = await inquirer.prompt([
```
amcgee, line 91: (minor) should we also prompt for `id` in the case this isn't a platform app?

cli/src/commands/publish.js
Birkbjo: I agree! That would be great, however I didn't want to make this dependent on adding support for that as well.

cli/src/commands/publish.js
Birkbjo: Can add that, I do have `implies: 'appId` if `file`-param is present though, so you would need to provide the `appId` from param, but I guess prompting for that also is better.

amcgee: Minor, if you think it's OK as-is for now that's fine - there might be cases where the app is built with the platform but no appId has been specified, in which case it would be nice to prompt (and, even more nice, to offer to save the prompt input to the d2.config.js file - though that's not as easy since it's a dynamic file.

amcgee:
@Birkbjo is this blocked by anything?  Ready to be reviewed or not yet?

amcgee:
Thought: maybe we should use the environment variable `D2_APP_HUB_API_KEY` or something a bit more explicit?  `D2_APIKEY` could be the api key for anything, we might have need of other api keys in the future?

Birkbjo:
> Thought: maybe we should use the environment variable `D2_APP_HUB_API_KEY` or something a bit more explicit? `D2_APIKEY` could be for anything, we might have need of other api keys in the future?

This is due to the built in resolution of env-vars to yargs params. Should I move away from using that in this case?

amcgee:
Yeah, I understand that, I don't think we should move away... I think it
might be nice to have a more explicit version (in addition) since we might
expect people to have the API key in their .bashprofile or something?
Maybe `const resolvedApiKey = apiKey /* from cli */ ||
process.env.D2_APP_HUB_API_KEY || prompt()` ... ?

cli/src/commands/publish.js
```diff
@@ -0,0 +1,269 @@
...
+            )
+            process.exit(1)
+        }
+    }
+
+    return appBundle
+}
+
+const promptForConfig = async params => {
+    if (process.env.CI && (!params.apikey || !params.minVersion)) {
```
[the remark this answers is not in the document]
Birkbjo: > @Birkbjo I think this will throw if the API key is passed as an env var in CI, right? We actually should prefer to use env vars in CI

Agree that we should prefer env-vars. But good catch, this would work with the previous env-var, as that would be resolved by yargs - but I totally forgot that this would not work when I changed the name.

cli/src/commands/publish.js
```diff
@@ -0,0 +1,275 @@
...
+        reporter.error('Prompt disabled in CI mode - missing apikey parameter.')
+        process.exit(1)
+    }
+
+    const responses = await inquirer.prompt([
+        {
+            type: 'input',
+            name: 'apikey',
+            message: 'App Hub API-key',
+            when: () => !params.apikey && !process.env.D2_APP_HUB_API_KEY,
```
amcgee, line 139:
```suggestion
            when: () => !apiKey,
```

cli/src/commands/publish.js
```diff
@@ -0,0 +1,275 @@
...
+            validate: v =>
+                !v || isValidServerVersion(v) ? true : 'Invalid server version',
+        },
+    ])
+
+    return {
+        ...params,
+        ...responses,
+    }
+}
```
amcgee, line 175:
```suggestion
    return {
        ...params,
        apikey: apiKey,
        ...responses,
    }
}
```

cli/src/commands/publish.js
```diff
@@ -0,0 +1,275 @@
...
+    const uploadAppUrl = constructUploadUrl(appBundle.id)
+
+    const resolvedApiKey =
+        publishConfig.apikey || process.env.D2_APP_HUB_API_KEY
+    const client = createClient({
+        baseUrl: publishConfig.baseUrl,
+        headers: {
+            'x-api-key': resolvedApiKey,
+        },
+    })
```
amcgee, line 190:
```suggestion
    const client = createClient({
        baseUrl: publishConfig.baseUrl,
        headers: {
            'x-api-key': publishConfig.apikey,
        },
    })
```

cli/src/commands/publish.js
```diff
@@ -0,0 +1,275 @@
...
+
+    try {
+        reporter.print(
+            `Uploading app bundle to ${publishConfig.baseUrl + uploadAppUrl}`
+        )
+        reporter.debug('Upload with version data', versionData)
+
+        await client.post(uploadAppUrl, formData, {
+            headers: formData.getHeaders(),
+            timeout: 30000, // Ensure we have enough time to upload a large zip file
```
amcgee, line 211:
```suggestion
            timeout: 300000, // Ensure we have enough time to upload a large zip file
```

For me on a mildy-fast internet connection I was sometimes having timeouts uploading an app, let's make this nice and long (5 minutes)

amcgee (commented):
Looking good!  I added some suggestions just to clean up the logic a touch

cli/src/commands/publish.js
mediremi: We should also ensure the app hub backend has a long timeout for that endpoint :thinking:

amcgee: @mediremi agreed, do you know what the timeout currently is?  It's at least more than 30 seconds (the previous value) since I was able to upload after bumping the client timeout

Birkbjo: As far as I can see we have not specified it anywhere, and the [docs](https://hapi.dev/api/?v=20.1.2#-routeoptionspayloadtimeout) says its a default value of 10 seconds, but it's obviously not? Maybe it's different for streamed payloads.

cli/src/commands/publish.js
varl: Add an issue to the CLI project in Jira so we don't forget?

cli/src/commands/publish.js
varl: The timeout on our load balancer in AWS is 60 seconds.

mediremi: Since we support bundles of up to 20MB maybe we should try to increase that to ~5mins?

varl: Sounds good to me. The docs on how to do so is here: https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-idle-timeout.html

We can do it using the AWS GUI as long as we don't destroy/recreate our environments that often.

We should document the AWS setup at some point as well so we know what customization we apply.

Birkbjo: @varl  that is different than timeout right? The timeout is for how long a request is allowed to take when uploading something AWS describes that timeout as `idle`: " If no data has been sent or received by the time that the idle timeout period elapses, the load balancer closes the connection." Data is still being sent (hopefully) when uploading takes longer than that, so to me that doesn't seem to be the same thing as we're talking about here?

mediremi (approved):

amcgee (approved):
Looks good, let's get this in!  🍾

cli/src/commands/publish.js
varl: I'm not sure, you may be correct.

> For each request that a client makes through a Classic Load Balancer, the load balancer maintains two connections. The front-end connection is between the client and the load balancer. The back-end connection is between the load balancer and a registered EC2 instance. The load balancer has a configured idle timeout period that applies to its connections.

This also stood out to me as relevant:

>To ensure that lengthy operations such as file uploads have time to complete, send at least 1 byte of data before each idle timeout period elapses, and increase the length of the idle timeout period as needed.

The way I interpret the docs is that if a client uploads a file, it is sent to the load balancer, and when it's complete the load balancer sends it to the backend. It implies that the data isn't just streamed through one connection to another.

If that's true, then the docs makes sense to me since if it takes longer to upload the file to the load balancer than it maintains the connection to the backend, it would close the connection it has to the backend before the file upload completes.

I may be misunderstanding the docs, and I cannot find other docs explaining it clearly. We pay for the Load Balancer on a "per GB transferred" basis too, but I don't know if that means anything. :shrug:

PR #537, opened 2021-03-17
fix: catch service worker unregistration rejection

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-155

### DIFF
--- a/shell/src/serviceWorker.js
+++ b/shell/src/serviceWorker.js
@@ -129,8 +129,15 @@ function checkValidServiceWorker(swUrl, config) {
 
 export function unregister() {
     if ('serviceWorker' in navigator) {
-        navigator.serviceWorker.ready.then(registration => {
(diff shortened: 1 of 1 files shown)


Mohammer5 (commented):
This is now logging all error.
Maybe it makes sense to only prevent the unhandled rejection for the specific error (and maybe only in development mode)?

amcgee (approved):
I think it's fine to catch all promise rejections here, we can fine-tune the error message or handling for different errors in the future.

I'm not sure this is the linked issue, actually - this error also happens when a serviceworker is registered on localhost - but it's not clear why it's happening here.  In any case, it shouldn't hurt to catch rejections.

Mohammer5:
@amcgee can be reproduced by switching between before- & after-platform-upgrade commits/branches with Firefox and the setting that cookies should be deleted when closing the browser

PR #538, opened 2021-03-19
feat(graphql): create babel plugin to convert gql to app-runtime queries /wo mutations

Mohammer5:
Fixes LIBS-158

:heavy_check_mark: Done
* Can convert queries (not mutations) from gql to app-runtime queries
* Adds support for the `*` field with the `__all` property (see tests)
* Tested with jest
* Make sure the plugin works when used with babel
* Add the plugin to the app platform
* Iteration 1: Support for queries, but not for mutations

:question: Open questions
* Do we need to add other fields, similar to the `__all` field? (I remember @amcgee mention `:owner`, but I've never heard of / used that field so far


amcgee:
Nice @Mohammer5 !  Regarding where this should live, I refactored things in the dynamic modules branch to store configs as separate packages ([here](https://github.com/dhis2/app-platform/tree/feat-dynamic-modules/configs))  which I think would be a good place for this - it should be its own package published on `npm` I think.

Mohammer5:
@amcgee I've created an extra workspace for the plugin so we can release it as its own package.
I've also clean the workspaces up a bit by moving them (except the example app) into a `packages` folder

amcgee:
>  I've created an extra workspace for the plugin so we can release it as its own package.
👍 

> I've also clean the workspaces up a bit by moving them (except the example app) into a packages folder
Let's not do that in this PR.  It will cause a lot of conflicts with other branches, and we actually chose *not* to use `packages` early on because @varl and I preferred the root-packages aesthetic.  I'm not opposed to switching to the more conventional approach, but we should do that in its own change.

For now in this PR let's keep the babel plugin in `config/<name>` to match #529 and we can restructure the package layout in a separate change

Mohammer5:
I've dropped the commit that moves the packages, so that should be ok again

Mohammer5:
I've added some questions that we need to discuss before I can proceed

Mohammer5:
@amcgee @varl I think this is ready for review, I've added it to the `cli` package and added a graphql query to the example app, which worked right out of the box

varl (changes requested):
This is **very** cool, but I don't understand the practical use case of this.

I've added a comment in the LIBS-158 issue to discuss, as it is off-topic to the PR itself, but adding a blocking review here until that's resolved.

amcgee:
Some implementation feedback I shared with @Mohammer5 -

- We have one logical query operation here (I'll call it `getResource`) which we should make explicit using [graphql aliasing](https://graphqlmastery.com/blog/how-to-use-graphql-aliases) - i.e. `smsCommands: getResource(...) { }` rather than `smsCommands(...) { }` - @Mohammer5 said he would implement this.
- We should consider the implications of using a single query operation (`getResource`) with variable type based on the Resource provided.  It would be more "graphql-esque" and also could directly talk to a graphql server if we use strongly-typed resources (i.e. `getSmsCommands(filter: 'xyz', page: 4) { id, dataSet { id, displayName } }`).  We could probably use a codemod to transform `getResource(resource: 'smsCommands')` into `getSmsCommands`, but if we could do this from the start it would be a bit cleaner.  The challenges here (1) collections vs single resources and (2) complex resource paths (i.e. `sms/inbound` would have to become `getInboundSMSs` or something).  As this is a bit complex, we should probably defer it for now and perform the transform later.

@varl brings up some very good points in LIBS-158, worth considering if this is the direction we want to go.  I will respond there, but a few quick thoughts:
1. In order to do our own static analysis or schema validation against our DataQuery objects, we would need to implement a lot of our own tooling.  Using the gql string format (which is conceptually almost identical to our data query format, by design) we can leverage the large gql tooling ecosystem even if we don't have a real graphql server for a long time
2. We need to move to strongly-typed query objects (i.e. `const query = new Query({ ... })` instead of `const query = {}`) anyway if we ever want to do static analysis on our existing format.  This would require some pretty fancy typescript types which of course could be done, but using a graphQL schema is probably easier and more portable to a future graphql BFF service.
3. Explicitly declaring query variables in a string format (which GraphQL queries support) allows us to do much better static analysis and code completion of those queries - with the function expressions we currently use for dynamic data queries, the functions are opaque and it's very difficult to figure out what's going on inside them without forcing typescript in our applications

All that being said, there are some reasonable concerns we should address with introducing this now _as a parallel option_, as it does bifurcate the ecosystem somewhat.  We should discuss those :-)

Mohammer5:
Copied from slack, said by @varl:

> austin commented on my remarks, and i accept those arguments. the only two points i have outstanding as reservations are:
> - is having two ways of doing queries worth the extra complexity for our users. how would our best practices surrounding this look? when do we recommend what?
> - i think we need mutation support for this to be useful and consistent in code. otherwise we end up with fetches using gql and mutations using object notation, which breaks the abstraction for the user.
>
> ref. mutations: to avoid this pr getting stale i support merging in mutation support in a different pr, but we need that before we recommend use this.
> ref. two ways of doing things: i am concerned about what happens if we do changes to the query objects in terms of the GQL layer and the GQL=>Query babel plugin.
> is our code coverage good enough to ensure that we don't break those additional layers when working on "lower" level Query?

The second point is not really a problem, we could, as you said, either add the functionality to the PR or before we announce that we've added the functionality

---

Before I talk about the first point, I'd like to talk about why I think that graphql is more than a nice addition to the app-runtime.

Our current situation is that we're quite understaffed (as least with developers, don't know about other departments), which is something our whole industry suffers from. We've introduced libraries/utilities (not exclusively for that purpose) that help us manage this gap, like the app-runtime, app-platform, cli, etc. One of the biggest issue with this approach is, that we have to maintain these ourselves. In some cases that's unavoidable, in other cases we could've leveraged on existing technologies out there and add only what we need to customize the utility to work for us. The fact, that we have our own data engine instead of using graphql can easily be explained: We don't have a graphql schema/server but an existing custom schema and api that can work with that schema.

Graphql has a very large user base (compared to our tools), lots of contributors and a ton of libraries and plugins to handle all kinds of scenarios that people encounter. So, I guess, we can all agree, that - if we had all the necessary infrastructure in place - using graphql would give us a lot of benefits that we currently miss. The reason why I said it'd be nice for us to follow that standard, is: The standard evolves and gets better without us having to do anything, which doesn't mean we don't have to contribute, but essentially we wouldn't to.

The way our api has been implemented is based on a very similar concept that graphql tries to solves as well, which is why we can translate a graphql query "so easily" to a app-runtime query.

I think that, **if we managed to somehow transition to graphql, we'd greatly benefit from it** for many reasons, some of which I've mentioned above. The question that arises (at least to me), is: **How do we get there?**

Here's what I thought might be a good way, which of course doesn't mean it's the best way and of course I'm open to change it entirely (if we decide to go for it):

1. We introduce the possibility to use graphql instead of data-engine queries
   * Under the hood these will be translated to data-engine queries
   *  We do not encourage to use these instead of the data-engine ones. Instead we'll try to play around with these in a limited scope, e. g. just one app, from there we can expand.
2. Once we've completed step 1 and think about it as having been successful, we open up the idea of using gql to other apps.
   * We could provide a codemod that transforms existing data-engine queries to gql queries
3. We add a graphql server that uses our schema, which would replace the babel plugin that's being added with this PR. The app-runtime would have to send these requests directly to the gql server.
   * By doing so, we'll allow our apps to leverage on existing functionality, like the apollo client, that can take away a lot of work that we're currently entangled with (e. g. caching, deduplication, etc)
   * The server would have auto-generated resolvers that query the actual api

At any point in time, we can "easily" revert changes with a codemod that transforms gql queries to data-engine queries, as that's what the babel plugin does as well. The following concern is still valid, though:

> ref. two ways of doing things: i am concerned about what happens if we do changes to the query objects in terms of the GQL layer and the GQL=>Query babel plugin.
> is our code coverage good enough to ensure that we don't break those additional layers when working on "lower" level Query?

The way I think about it is: There are very few things that can't be done with graphql. One thing that comes to my mind is the json-patch functionality. I'm not sure how we would have to handle this, but a quick google search indicates that we're not the only ones with this question and it seems there are solutions out there. Worst case scenario: We'll have to implement our own [directives](https://graphql.org/learn/queries/#directives), which are quite powerful but should be the last resort (if we really need an escape hatch). [See what neo4j did](https://neo4j.com/docs/graphql-manual/current/type-definitions/cypher/) to allow using the cypher query language in the graphql schema

Mohammer5:
I suppose we can close this one. Graphql is a topic that will take more time and thinking

PR #539, opened 2021-03-23
chore(example app): add missing dep in example app

Mohammer5:
Can't run the build script of the example app without this dependency


amcgee:
@Mohammer5 I don't think this is correct... for one thing, this PR has a lot of other things in it which probably should stay in #538 .  For another, the example app definitely should not have a dependency on `app-adapter` - it is a dependency of `app-shell`.  Most likely the issue you saw was that `app-adapter` hadn't yet been built locally - if you run `yarn start` at the repo root it will build the adapter before running the example and should always work.

amcgee:
Feel free to re-open if this is incorrect.

PR #540, opened 2021-03-24
fix: respect --cwd argument when generating/extracting translations

mediremi:
Currently, when generating `i18n` strings (either as a result of `d2-app-scripts start` or `d2-app-scripts extract`), a message in the format below is logged:

```
Writing N language strings to .//home/user/some/path/app-being-developed/i18n/en.pot...
```

This PR ensures that the path to `en.pot` logged is relative to the current working directory, resulting in the more consice:

```
Writing N language strings to ./en.pot...
```

### DIFF
--- a/cli/src/commands/i18n.js
+++ b/cli/src/commands/i18n.js
@@ -1,32 +1,25 @@
 const { namespace, reporter } = require('@dhis2/cli-helpers-engine')
 const i18n = require('../lib/i18n')
+const makePaths = require('../lib/paths')
 
 const generate = {
     description:
         'Generate JSON files compatible with i18next from po/pot files',
     builder: {
-        path: {
-            alias: 'p',
-            description:
-                'directory path to find .po/.pot files and convert to JSON',
-            default: './i18n/',
-        },
-        output: {
-            alias: 'o',
-            description: 'Output directory to place converted JSON files.',
-            default: './src/locales/',
-        },
         namespace: {
             alias: 'n',
             description: 'Namespace for app locale separation',
             default: 'default',
         },
     },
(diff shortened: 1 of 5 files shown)


amcgee (approved):
LGTM

i see there's actually an existing bug here - the CLI commands accept a `--cwd` argument, so `process.cwd()` is NOT always guaranteed to be the proper cwd.  This function should probably accept `paths` which includes the `cwd` arg (defaulting to `process.cwd()`) and use that instead - as is currently written, this change looks correct.  Feel free to fix the larger issue if desired, but if we want to defer the bugfix (likely also in `generate` as both were ported from https://github.com/dhis2/d2-i18n-generate / https://github.com/dhis2/d2-i18n-extract) we just can create an issue to follow up.

mediremi:
I'll update both `extract` and `generate` so that they use `paths.cwd` instead of `process.cwd()` :+1:

cli/src/commands/i18n.js
```diff
@@ -1,32 +1,25 @@
...
-            alias: 'p',
-            description:
-                'directory path to find .po/.pot files and convert to JSON',
-            default: './i18n/',
-        },
-        output: {
-            alias: 'o',
-            description: 'Output directory to place converted JSON files.',
-            default: './src/locales/',
-        },
```
amcgee, line 18: This makes this a breaking change, unfortunately - we need to keep the `path` and `output` options for backwards compatibility but fall back to `paths.i18nStrings` and `paths.i18nLocales` below

amcgee (changes requested):

amcgee (approved):

PR #549, opened 2021-04-21
fix(app-adapter): align `Alerts` component with new alerts-service implementation

HendrikThePendric:
This PR is in draft because it depends on https://github.com/dhis2/app-runtime/pull/820. Once that PR is merged we should upgrade to its released version. I tested locally with the unreleased version of https://github.com/dhis2/app-runtime/pull/820 and everything was working perfectly.

### TODO
- [x] Upgrade `app-platform` dependency once https://github.com/dhis2/app-runtime/pull/820 is merged


varl (approved):

examples/simple-app/yarn.lock
```diff
@@ -1,15218 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ismay, line 1: Should we not keep this file?

ismay (approved):
Looks good, just the removed yarn.lock I'm not sure about.

examples/simple-app/yarn.lock
varl: No, since `simple-app` is registered as a workspace it should not have its own `yarn.lock` file in the first place: https://github.com/dhis2/app-platform/blob/master/package.json#L11

ismay: Ah ok. Well seems good to merge then 👍

PR #552, opened 2021-05-05
fix: set default value for all strings

jenniferarnesen:
A default value will be set for every string. if no default value is specifically provided, then the key will be used.  The end result is that the generated `en.pot` file contains translations for every string. This pot file is then accepted by transifex, allowing plurals to be properly translated. (after much trial and error together with Phil, we found that Transifex will not parse the imported pot file if only some of the strings have values)

d2-ui/interpretations currently uses this modified version of the extract.js script and it works as expected: https://github.com/dhis2/d2-ui/blob/master/packages/interpretations/scripts/extract.js

Would it be possible to release a beta to be tested with the DV app?

Here is an example of how plurals would look in code:
```
const replies = i18n.t(
            '{{count}} replies',
            {
                count: repliedBy.length,
                defaultValue: '{{count}} reply',
                defaultValue_plural: '{{count}} replies'
            });
```
(https://github.com/dhis2/d2-ui/blob/master/packages/interpretations/src/components/Interpretation/Replies.js#L73)

### DIFF
--- a/cli/src/lib/i18n/extract.js
+++ b/cli/src/lib/i18n/extract.js
@@ -30,7 +30,7 @@ const extract = async ({ input, output, paths }) => {
         keySeparator: false,
         sort: true,
         defaultValue: (lng, ns, key, options) =>
-            options.defaultValue ? options.defaultValue : key
+            options.defaultValue ? options.defaultValue : key,
     })
     /* eslint-enable max-params */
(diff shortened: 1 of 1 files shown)


amcgee (approved):
LGTM, nice find @jenniferarnesen !

PR #561, opened 2021-05-18
fix: add file extension to default entrypoints in d2.config

KaiVandivier:
This fixes an issue when using cjs `require`s from a d2 lib package. Also added to the default app `d2.config` for consistency


amcgee (approved):
LGTM!  We should probably update the [validation logic](https://github.com/dhis2/app-platform/blob/master/cli/src/lib/validators/validatePackageExports.js) to check for an extension in d2.config.js too

PR #576, opened 2021-06-22
chore(deps): update ui to 6.10.0

varl:


amcgee (approved):
Assuming we don't want to let `@dhis2/ui` bake for too much longer, I think this should be fine!  There could be issues with duplicate @dhis2/ui dependencies if the app depends on their own copy and yarn does its weirdness, but that's unavoidable anyway.

varl:
Alright, I'll merge this once UI has baked a bit.

amcgee:
@varl baking complete, let's get this out of the oven

PR #577, opened 2021-06-30
fix: rename 'update' test command handler parameter to 'updateSnapshot'

mediremi:
The [`updateSnapshot` parameter for the `test` command](https://github.com/dhis2/app-platform/blob/4e704c74bc5f02774f360fe65d8a684749affc02/cli/src/commands/test.js#L77-L82) was incorrectly referred to as `update` in its handler, hence it was impossible to set it to `true`.

### DIFF
--- a/cli/src/commands/test.js
+++ b/cli/src/commands/test.js
@@ -9,7 +9,7 @@ const handler = async ({
     verbose,
     cwd,
     testRegex,
-    update,
+    updateSnapshot,
     coverage,
     watch,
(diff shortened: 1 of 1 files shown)


amcgee (approved):
Nice catch, lgtm!  Not sure why publish step is failing... @varl

amcgee:
Shoot... might be a problem with the publish command here?  Maybe it's a problem with `cli-utils release`?

PR #578, opened 2021-06-30
refactor: introduce pack command

varl:


varl:
This is now non-breaking so I'll merge it to alpha and test it out in the pipeline.

PR #581, opened 2021-06-30
chore(release): beta to master

varl:
Follow-ups in: https://github.com/dhis2/notes/discussions/293


cli/src/commands/pack.js
```diff
@@ -0,0 +1,81 @@
...
+    if (config.type === 'app') {
+        inputPath = path.resolve(cwd, folder ? folder : paths.buildAppOutput)
+
+        outputPath = path.resolve(
+            cwd,
+            destination ? destination : paths.buildAppBundleOutput,
+            filename ? filename : paths.buildAppBundleFile
+        )
+    } else {
+        inputPath = path.resolve(cwd, folder ? folder : '.')
```
amcgee, line 45: `'.'` shouldn't be the default for library builds right?  The build output is at `./build/es` or `./build/cjs` (or just `./build`) by default - should be able to get this from `paths`

cli/src/commands/pack.js
```diff
@@ -0,0 +1,81 @@
...
+
+exports.describe = 'Create archive from the build.'
+
+exports.builder = yargs =>
+    yargs
+        .positional('folder', {
+            describe: 'The folder to pack relative to cwd.',
+            type: 'string',
+            defaultDescription: '.',
+        })
```
amcgee, line 17: `folder` seems a bit ambiguous here - could mean either input or output (and I would actually assume output on first glance).  Maybe `sourceDirectory` or `inputDirectory` and then the complementary `destinationDirectory` or `outputDirectory` or `targetDirectory`?

cli/src/commands/pack.js
```diff
@@ -0,0 +1,81 @@
...
+exports.command = 'pack [folder]'
+
+exports.describe = 'Create archive from the build.'
+
+exports.builder = yargs =>
+    yargs
+        .positional('folder', {
+            describe: 'The folder to pack relative to cwd.',
+            type: 'string',
+            defaultDescription: '.',
```
amcgee, line 16: `'.'` might not be the correct default?

cli/src/commands/pack.js
```diff
@@ -0,0 +1,81 @@
...
+        )
+    }
+
+    // for scoped names in package.json
+    const clean = str => str.replace(/@/, '').replace(/\//, '-')
+
+    // replace placeholder within names defined in lib/paths.js
+    const archivePath = outputPath
+        .replace(/{name}/, clean(config.name))
+        .replace(/{version}/, config.version)
```
amcgee, line 70: Maybe we should extract this logic to the `makeBundle` helper?

cli/src/commands/publish.js
```diff
@@ -170,17 +158,47 @@ const promptForConfig = async params => {
...
+    const appConfig = parseConfig(paths)
 
-    const appBundle = resolveBundle(cwd, publishConfig)
+    let publishConfig
+
+    if (process.env.CI) {
+        publishConfig = params
+    } else {
+        publishConfig = await promptForConfig(params)
+    }
```
amcgee, line 175: Nice!

cli/src/commands/publish.js
```diff
@@ -229,7 +247,7 @@ const command = {
     builder: yargs =>
         yargs.options({
             apikey: {
-                alias: 'k',
+                alias: ['k', 'app-hub-api-key', 'app-hub-token', 'token'],
```
amcgee, line 250: Do all of these show up in the help text?

cli/src/lib/makeBundle.js
```diff
@@ -30,7 +30,31 @@ module.exports = (inDir, outFile) => {
         })
 
         archive.pipe(output)
-        archive.directory(inDir, false)
+
+        // avoid packing the created file stream
```
amcgee, line 34: Just out of curiosity, what does this mean exactly?

amcgee (commented):
Looks great!  Some minor comments / questions

cli/src/lib/makeBundle.js
varl: It means that if we just pass in the `.directory(inDir)` and create the file stream in `inDir/bundle.zip`, then the `bundle.zip` is going to end up inside the zip (https://github.com/archiverjs/node-archiver/issues/319).

This is not much of a problem with the defaults for an app, but for a lib it can be.

cli/src/commands/publish.js
varl:
```
d2-app-scripts publish

Deploy the built application to a specific DHIS2 instance

Global Options:
  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]
  --verbose      Enable verbose messages                               [boolean]
  --debug        Enable debug messages                                 [boolean]
  --quiet        Enable quiet mode                                     [boolean]
  --config       Path to JSON config file

Options:
  --cwd                                     working directory to use (defaults
                                            to cwd)
  --apikey, -k, --app-hub-api-key,          The API-key to use for
  --app-hub-token, --token                  authentication              [string]
  --channel, -c                             The channel to publish the
                                            app-version to   [default: "stable"]
  --baseUrl, -b                             The base-url of the App Hub instance
                                             [default: "https://apps.dhis2.org"]
  --minDHIS2Version                         The minimum version of DHIS2 the app
                                            supports                    [string]
  --maxDHIS2Version                         The maximum version of DHIS2 the app
                                            supports                    [string]
  --appId                                   Only used with --file option. The
                                            App Hub ID for the App to publish to
                                                                        [string]
  --file                                    Path to the file to upload. This
                                            skips automatic resolution of the
                                            built app and uses this file-path to
                                            upload
  --file-version                            Only used with --file option. The
                                            semantic version of the app uploaded
                                                                        [string]
```

They do. We can [hide](https://github.com/yargs/yargs/blob/master/docs/api.md#hidekey) the keys we don't want to promote.

cli/src/commands/pack.js
varl: OK, we use `source` `destination`/`dest` in other places so I'll align with that for consistency.

cli/src/commands/pack.js
varl: For clarity, it is simply a **description** of the internal default, it will never get set to `.`, which e.g. `default: '.'` would do.

I'll remove it. I chose `.` instead of printing the full `cwd` because it's logically equivalent even if it's not technically true. Maybe it doesn't help the user at all.

cli/src/commands/pack.js
varl: Done.

cli/src/commands/pack.js
varl: Yeah, on second thought, we don't need `pack` to support libraries right now.

cli/src/commands/pack.js
```diff
@@ -27,30 +27,32 @@ exports.builder = yargs =>
...
-    const config = fs.readJsonSync(paths.buildAppConfigJson)
+    const config = fs.readJsonSync(paths.buildAppConfigJson, { throws: false })
 
-    let inputPath, outputPath
-    if (config.type === 'app') {
-        inputPath = path.resolve(cwd, folder ? folder : paths.buildAppOutput)
+    if (!config) {
+        // we may be dealing with a library or a unbuilt app
+        // load the d2.config.js file from the project and check
+        const baseConfig = require(paths.config)
```
amcgee, line 38: I think this will throw if the config file doesn't exist (i.e. if used on build output from a non-platform app) which I think would be a really useful feature (packing a directory is currently a big pain because you need to make absolutely sure you make a flat zip etc.  I'll add a check for file existence here.

amcgee (approved):
🎉 
I'm just going to fix conflicts, add a file existence check, and merge this

PR #582, opened 2021-07-01
feat(adapter): smarter error boundaries

KaiVandivier:
Adds an error boundary around the app to catch errors there, and enables the top-most error boundary to check for and apply service worker updates when 'refresh' is clicked

### DIFF
--- a/adapter/src/components/ErrorBoundary.js
+++ b/adapter/src/components/ErrorBoundary.js
@@ -1,7 +1,8 @@
 import i18n from '@dhis2/d2-i18n'
+import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
-import styles from './styles/FatalErrorBoundary.style'
+import styles from './styles/ErrorBoundary.style'
 
 const translatedErrorHeading = i18n.t(
     'An error occurred in the DHIS2 application.'
@@ -12,7 +13,7 @@ const replaceNewlinesWithBreaks = text =>
         .split('\n')
         .reduce((out, line, i) => [...out, line, <br key={i} />], [])
 
-export class FatalErrorBoundary extends Component {
+export class ErrorBoundary extends Component {
     constructor(props) {
         super(props)
         this.state = {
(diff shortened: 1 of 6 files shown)


adapter/src/components/ErrorBoundary.js
```diff
@@ -12,7 +13,7 @@ const replaceNewlinesWithBreaks = text =>
         .split('\n')
         .reduce((out, line, i) => [...out, line, <br key={i} />], [])
 
-export class FatalErrorBoundary extends Component {
+export class ErrorBoundary extends Component {
```
KaiVandivier, line 16: FatalErrorBoundary is repurposed into a reusable error boundary with the option of making it fullscreen or not, and configuring what happens on clicking 'refresh'

adapter/src/components/styles/ErrorBoundary.style.js
```diff
@@ -8,11 +8,8 @@ const bgColor = '#F4F6F8',
 
 export default css`
     .mask {
-        position: absolute;
-        top: 0px;
-        left: 0px;
-        right: 0px;
-        bottom: 0px;
+        height: 100%;
+        width: 100%;
```
KaiVandivier, line 12: By default, the mask will fill up its container

adapter/src/components/styles/ErrorBoundary.style.js
```diff
@@ -29,6 +26,14 @@ export default css`
...
     }
 
+    .fullscreen {
+        position: absolute;
+        top: 0px;
+        left: 0px;
+        right: 0px;
+        bottom: 0px;
+    }
+
```
KaiVandivier, line 36: If the `fullscreen` prop is added to the `ErrorBoundary`, it will take up the whole screen instead

examples/pwa-app/package.json
```diff
@@ -10,7 +10,7 @@
         "test": "d2-app-scripts test",
         "deploy": "d2-app-scripts deploy",
         "serve": "serve -l 5500 build/app",
-        "demo": "yarn build --force --verbose && yarn serve"
+        "demo": "yarn build --force --debug && yarn serve"
```
KaiVandivier, line 13: Just a change to use the correct flag here... did this change recently?

mediremi (approved):

PR #590, opened 2021-07-12
feat: support new OfflineProvider API for new state management

KaiVandivier:
Supports new API that new state management in the app-runtime uses: https://github.com/dhis2/app-runtime/pull/901


amcgee:
See comments on dhis2/app-runtime#901 - I don't know that it's necessary to construct the store outside the provider?

KaiVandivier:
Agreed -- closing this!

PR #592, opened 2021-07-16
feat: show error message in fatal error boundary and allow copying stack trace to clipboard

mediremi:
Improve error boundary by:
- Including error message in technical details drawer
- Add button to copy error message and stack trace to clipboard
- Remove min-width on error boundary mask to better support non-desktop devices
- Using `<button>` instead of `<div>` for drawer toggling

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/adapter/i18n/en.pot
+++ b/adapter/i18n/en.pot
@@ -5,11 +5,14 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-07-14T21:34:51.462Z\n"
-"PO-Revision-Date: 2021-07-14T21:34:51.462Z\n"
+"POT-Creation-Date: 2021-07-16T13:22:12.456Z\n"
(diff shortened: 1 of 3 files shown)


cooper-joe:
LGTM 👍 

As an aside, anyone know what's going on with the font-rendering here? The character spacing looks to be quite jittery and random, with some of the characters being mashed together. @mediremi - what browser/OS were the screenshots captured from?
letter-spacing

cooper-joe (approved):

mediremi:
Turned out the weird kerning/hinting was due to [some Linux distros (including the one I use) bundling broken ChromeOS font configs that force full hinting for Roboto](https://bugzilla.mozilla.org/show_bug.cgi?id=1481545). Chrome on those distros ignores this broken config but Firefox does not.

I've removed the config file from my system and kerning is as it should be:
image

amcgee:
@mediremi nice work, looks good!

@cooper-joe should the non-mono text and button maybe be outside the code block?

Future enhancement - we should be able to use source maps to show the source stack trace rather than the minified one - CRA does this.  We'd want to make sure we're resilient to

cooper-joe:
> @cooper-joe should the non-mono text and button maybe be outside the code block?

Yeah, that's a good idea I think. It should still be within the expand/collapse area, but it makes sense to put it outside the codeblock.

PR #594, opened 2021-07-21
refactor: export subpaths from pwa package

KaiVandivier:
Currently not working with this set up -- Adapter package can build, but apps throw an error when built:

```sh
 Failed to compile.

 /Users/kai/dev/dhis2/app-platform/adapter/build/es/index.js
Cannot find module: '@dhis2/pwa/browser'. Make sure this package is installed.
```

When running an app with `yarn start`, nothing appears and an error '`Uncaught SyntaxError: expected expression, got '!'`' is logged

There's also an eslint error: `Unable to resolve path to module '@dhis2/pwa/sw'. eslint(import/no-unresolved)`.

Doesn't work with `file:...` resolutions or workspace resolution


mediremi:
One reason why this might not work is that according to `yarn why webpack` we are using `webpack@4.44.2` but support for the exports field was added to webpack 5 (see comments [1](https://github.com/webpack/webpack/issues/10555#issuecomment-599096040), [2](https://github.com/webpack/webpack/issues/9509#issuecomment-648908587) and [this PR](https://github.com/webpack/webpack/pull/10953)).

Another issue (which shouldn't affect this PR but might affect other libs like the runtime) is that [Typescript does not support package exports](https://github.com/microsoft/TypeScript/issues/33079).

KaiVandivier:
Ohhh interesting - I had actually come across that in my googling, and even run `yarn why` but I only noticed the top instance of it at 5.41.1 🤦‍♂️

KaiVandivier:
I also wonder if yarn workspaces has something to do with it?

PR #597, opened 2021-07-26
fix(cli): Support multiple library entrypoints

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-202

This PR adds support for multiple library entrypoints defined in d2.config.js, which enables the use of the `exports` field of `package.json` for conditional exports. This is particularly useful for libraries which define different entrypoints for web worker/service worker contexts.


cli/src/lib/compiler/entrypoints.js
```diff
@@ -0,0 +1,72 @@
...
+        const msg = `Could not resolve entrypoint ${chalk.bold(entrypoint)}`
+        reporter.error(msg)
+        throw new Error(msg)
+    }
+}
+
+exports.verifyEntrypoints = ({
+    config,
+    paths,
+    resolveModule = require.resolve,
```
mediremi, line 27: This was required as it is not currently possible to mock `require.resolve` using jest

cli/src/lib/validators/validatePackageExports.js
```diff
@@ -120,7 +194,7 @@ module.exports.validatePackageExports = async (
             return false
         }
         await fixPackage(pkg, expectedPackage, { paths })
-        valid = true
+        return true
     }
-    return valid
+    return !!valid
```
mediremi, line 199: The use of bitwise and assignment (`&=`) means that `valid` will be `1` instead of `true` and `0` instead of `false` and hence we must coerce `valid` to a boolean.

While this could be fixed by using logical and assignment (`&&=`), that operator is only supported in node v16+.

KaiVandivier:
Nice 🎉 except for the little edge case mentioned above, this looks good, and nice tests 👍 

I merged these changes into the [pwa branch](https://github.com/dhis2/app-platform/pull/594) to see if it would work, and unfortunately it does not 🤔 I can't tell why not yet

cli/src/lib/validators/validatePackageExports.js
```diff
@@ -20,94 +20,168 @@ const fixPackage = async (pkg, expectedPackage, { paths }) => {
...
+
+const validateMultipleEntrypoints = (pkg, { config, paths }) => {
+    const expectedExports = getExpectedExports(config.entryPoints.lib, paths)
+    const expectedPackage = {
+        exports: expectedExports,
+    }
+    if (expectedExports['.']) {
+        expectedPackage.module = expectedExports['.'].import
+        expectedPackage.main = expectedExports['.'].require
+    }
```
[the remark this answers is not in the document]
mediremi: Oops :sweat_smile:

cli/src/lib/validators/validatePackageExports.js
```diff
@@ -115,8 +115,12 @@ const validateMultipleEntrypoints = (pkg, { config, paths }) => {
...
     }
     if (expectedExports['.']) {
-        expectedPackage.module = expectedExports['.'].import
-        expectedPackage.main = expectedExports['.'].require
+        if (expectedExports['.'].import) {
+            expectedPackage.module = expectedExports['.'].import
+        }
+        if (expectedExports['.'].require) {
+            expectedPackage.main = expectedExports['.'].require
+        }
```
KaiVandivier, line 123: This is better now that it won't erase `module` and `main` properties in `package.json`, but it also won't validate `main` and `module` unless `exports['.']` directly has `.import` and `.require` properties...

Is it important to have `main` and `module` properties? If so, you might want to follow down the `exports` properties recursively to find `import` and `require` values; if not, this is fine as it is

KaiVandivier: Actually I may be misinterpreting this - taking another look

mediremi: > Is it important to have main and module properties?

`main` is for node versions 12 and below (which do not support `exports`), while `module` was [a proposal](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md#typical-usage) that node does not support so I think it's fine to ignore both when their 'correct' values are ambiguous.

KaiVandivier: Okay cool :)

KaiVandivier (approved):

PR #605, opened 2021-08-11
feat(cli): add jestConfig option to test script

mediremi:
Usage: `d2-app-scripts test --jest-config ../path/to/jest.config.js`

### DIFF
--- a/cli/src/commands/test.js
+++ b/cli/src/commands/test.js
@@ -1,9 +1,20 @@
+const path = require('path')
 const { reporter } = require('@dhis2/cli-helpers-engine')
 const fs = require('fs-extra')
 const { runCLI } = require('jest-cli')
 const exitOnCatch = require('../lib/exitOnCatch')
 const loadEnvFiles = require('../lib/loadEnvFiles')
 const makePaths = require('../lib/paths')
+
+const getAppJestConfig = ({ jestConfigPath, paths }) => {
+    if (jestConfigPath) {
+        return require(path.resolve(paths.base, jestConfigPath))
+    } else if (fs.existsSync(paths.jestConfig)) {
+        return require(paths.jestConfig)
+    } else {
+        return {}
+    }
+}
 
 const handler = async ({
     verbose,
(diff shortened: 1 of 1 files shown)


cli/src/commands/test.js
```diff
@@ -1,10 +1,21 @@
...
 const { reporter } = require('@dhis2/cli-helpers-engine')
 const fs = require('fs-extra')
 const { runCLI } = require('jest-cli')
 const exitOnCatch = require('../lib/exitOnCatch')
 const loadEnvFiles = require('../lib/loadEnvFiles')
 const makePaths = require('../lib/paths')
 
+const getAppJestConfig = ({ jestConfigPath, paths }) => {
+    if (jestConfigPath) {
+        return require(path.join(paths.base, jestConfigPath))
```
varl, line 11: We should use `path.resolve` instead of `path.join` here so that both relative and absolute paths to the jest config works.

`paths.base` is always an absolute path: https://github.com/dhis2/app-platform/blob/master/cli/src/lib/paths.js#L10 so this should hold true:

`path.join`:

```js
// relative path: OK
> path.join('/path/to/app/project/component', '../jest.config.js')
'/path/to/app/project/jest.config.js'

// absolute path: ERR
> path.join('/path/to/app/project/component', '/path/to/project/component/jest.config.js')
'/path/to/app/project/component/path/to/project/component/jest.config.js'
```

`path.resolve`:

```js
// relative path: OK
> path.resolve('/path/to/app/project/component', '../jest.config.js')
'/path/to/app/project/jest.config.js'

// absolute path: OK
> path.resolve('/path/to/app/project/component', '/path/to/project/component/jest.config.js')
'/path/to/project/component/jest.config.js'
```

varl (changes requested):
I don't think an absolute path to `--jest-config` will work as is, added a comment with more detail.

cli/src/commands/test.js
mediremi: oops :sweat_smile:

varl (approved):

PR #612, opened 2021-08-19
feat: allow RegEx for filter patterns in d2.config.js

KaiVandivier:
Allows developers to write filter patterns as RegExes or strings for `config.pwa.caching.patternsToOmit`

(In the future, if we use a custom build script to compile the service worker in production instead of using CRA's SW compiler, these patterns can be injected straight into the service worker without having to pass them as env vars)


docs/pwa/pwa.md
```diff
@@ -13,7 +13,7 @@ You can opt in to PWA features using options in `d2.config.json`. Here are the o
...
+| `pwa.caching.patternsToOmit`            | Array of RegExps or Strings                                         | A list of URL patterns to omit from the default caching strategies. Strings will be converted to RegExes using `new RegExp(str)` to test URLs. If a URL matches one of these patterns, that request will not be cached by the default caching strategies. Note that _all requests are cached during recording mode_, regardless of this setting. When choosing these URL filters, note that it is better to _cache too many things_ than to risk _not caching an important part of the app shell_, so choose your filter patterns accordingly.                   |
```
amcgee, line 16: @KaiVandivier is the conversion of strings to regex done in CRA?

amcgee (approved):
Docs question, code looks good

docs/pwa/pwa.md
KaiVandivier: It happens in the service worker: https://github.com/dhis2/app-platform/blob/f8e18234c0dfbfd13b57e78879475873ec72f05b/pwa/src/service-worker/utils.js#L46-L48

KaiVandivier: Now that I think about it, I should only convert those patterns from strings to regex once - I'll add that to a related PR

KaiVandivier: Overview of what's happening currently: `d2.config.js` has strings or Regexes => CLI converts any regexes to strings so all the patterns are strings => all patterns passed as strings as env vars => SW parses env vars as strings, then uses `new RegExp(patternString)` to test URLs

KaiVandivier: > Now that I think about it, I should only convert those patterns from strings to regex once - I'll add that to a related PR

Added to [this PR](https://github.com/dhis2/app-platform/pull/611)

PR #614, opened 2021-08-23
fix: reopen indexedDB if global state resets

KaiVandivier:
This fixes the (seemingly) intermittent `cannot read property 'put' of undefined` issue at the end of recordings: apparently, [service workers will 'sleep'/terminate when they go idle](https://web.dev/service-worker-mindset/#watch-out-for-global-state), at which point they lose their global state (but this doesn't happen while the Chrome dev tools are open, so errors can slip by in development).  IndexedDB is reopened on demand if necessary to restore the `self.dbPromise` state.

### DIFF
--- a/pwa/src/service-worker/service-worker.js
+++ b/pwa/src/service-worker/service-worker.js
@@ -37,11 +37,8 @@ export function setUpServiceWorker() {
     // TODO: control with env var
     self.__WB_DISABLE_DEV_LOGS = true
 
-    // Globals
+    // Globals (Note: global state resets each time SW goes idle)
 
-    // Will be populated upon activation with a promise that accesses the
-    // recorded sections IndexedDB using the `idb` library - see `createDB()`
-    self.dbPromise
     // Tracks recording states for multiple clients to handle multiple windows
     // recording simultaneously
     self.clientRecordingStates = {}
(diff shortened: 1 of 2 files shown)


pwa/src/service-worker/service-worker.js
```diff
@@ -37,11 +37,8 @@ export function setUpServiceWorker() {
...
 
-    // Globals
+    // Globals (Note: global state resets each time SW goes idle)
 
-    // Will be populated upon activation with a promise that accesses the
-    // recorded sections IndexedDB using the `idb` library - see `createDB()`
-    self.dbPromise
     // Tracks recording states for multiple clients to handle multiple windows
     // recording simultaneously
     self.clientRecordingStates = {}
```
mediremi, line 44: Should we store `clientRecordingStates` in session/local storage or is this state short lived?

KaiVandivier: I left this as-is because it will only be populated with content for the duration of a recording, which is pretty brief, so it's fine to reset it to `{}` on termination/startup, I believe.

That said, in the testing session last week, [Hendrik encountered an issue](https://dhis2.slack.com/archives/CJF3MDK8Q/p1629373941003500) that seemed to be related to `clientRecordingStates`. I haven't been able to replicate it, and it's possible those error messages were logged in the process of other things failing

mediremi (approved):

PR #615, opened 2021-08-23
fix: simplify message payload to be compatible with firefox

KaiVandivier:
Firefox can't handle any special objects in the `postMessage` API between service workers and clients, which causes those messages to go quietly unsent, which sometimes manifests as failed recordings getting stuck under a loading mask. This fix simplifies the payload for recording error messages so they work on firefox


mediremi (approved):

mediremi:
> causes those messages to go quietly unsent

:scream:

PR #616, opened 2021-08-24
feat: bump jest to v27

mediremi:
Jest 27 introduced modern fake timers, which allow us to better mock timers as well as mock the system date/time.


varl:
Did we land on this being a breaking change or not ?

mediremi:
I think we said we don't a new major version as by default `yarn upgrade` does not upgrade major versions.

varl:
Node 13 reached EOL in 2020-06-01 so we don't need to bump a major here I think.

We do signal that we require Node >=12, and the ">" is slightly problematic here.

I'll add a comment to the decision in meeting notes about why we aren't releasing this as a breaking change.

varl (approved):

PR #622, opened 2021-08-25
feat: bump jest to v27

mediremi:
Bump Jest to version 27, but this time as a breaking change.


varl:
`[breaking change]` is the wrong format.

The correct format is:

```
feat: allow provided config object to extend other configs

BREAKING CHANGE: `extends` key in config file is now used for extending other config files
```

Example from here: https://www.conventionalcommits.org/en/v1.0.0/#specification

mediremi:
Oops I thought it would be triggered as long as the string `breaking change` was present for some reason

PR #633, opened 2021-09-02
fix: import index css file to apply global styles

HendrikThePendric:
This is a follow-up on #631 where we applied a CSS fix for layering issues. The fix was correct, but the CSS rules weren't being applied to the app because the `index.css` file was not being imported.

So we are now importing the file and the styles from `index.css` are being applied to the app. However, this file already included some style rules, and since they weren't being applied before (due to the file not being imported) it is safer to remove them, which I've done in 4376038.


varl:
Good catch on an "oops I was sure that was imported". :sweat:

varl (approved):

PR #635, opened 2021-09-03
feat(cli): instance proxy server

mediremi:
Introduces a proxy server that can be used to test remote instances without relying on cross-site cookies.

Usage:

```
d2-app-scripts start --proxy https://debug.dhis2.org/dev
```

### DIFF
--- a/adapter/src/components/LoginModal.js
+++ b/adapter/src/components/LoginModal.js
@@ -11,10 +11,11 @@ import i18n from '../locales'
 import { post } from '../utils/api'
 
 const staticUrl = process.env.REACT_APP_DHIS2_BASE_URL
+const defaultUrl = process.env.REACT_APP_DHIS2_DEFAULT_BASE_URL
 
 export const LoginModal = () => {
     const [server, setServer] = useState(
-        staticUrl || window.localStorage.DHIS2_BASE_URL || ''
+        staticUrl || defaultUrl || window.localStorage.DHIS2_BASE_URL || ''
     )
     const [username, setUsername] = useState('')
     const [password, setPassword] = useState('')
--- a/adapter/src/components/LoginModal.js
+++ b/adapter/src/components/LoginModal.js
@@ -11,11 +11,10 @@ import i18n from '../locales'
 import { post } from '../utils/api'
 
 const staticUrl = process.env.REACT_APP_DHIS2_BASE_URL
-const defaultUrl = process.env.REACT_APP_DHIS2_DEFAULT_BASE_URL
 
 export const LoginModal = () => {
     const [server, setServer] = useState(
-        staticUrl || defaultUrl || window.localStorage.DHIS2_BASE_URL || ''
+        staticUrl || window.localStorage.DHIS2_BASE_URL || ''
     )
     const [username, setUsername] = useState('')
     const [password, setPassword] = useState('')
--- a/cli/src/lib/proxy.js
+++ b/cli/src/lib/proxy.js
@@ -100,7 +100,10 @@ exports = module.exports = ({ target, baseUrl, port, shellPort }) => {
             proxyRes.headers['set-cookie'] = sc.map(stripCookieSecure)
         }
 
-        if (proxyRes.headers['content-type']?.includes('application/json')) {
+        if (
+            proxyRes.headers['content-type'] &&
+            proxyRes.headers['content-type'].includes('application/json')
+        ) {
             transformProxyResponse(res, proxyRes, body => {
                 if (body) {
                     return transformJsonResponse(body, {
--- a/cli/src/lib/proxy.js
+++ b/cli/src/lib/proxy.js
@@ -1,6 +1,8 @@
 const url = require('url')
 const { reporter } = require('@dhis2/cli-helpers-engine')
 const httpProxy = require('http-proxy')
+const _ = require('lodash')
+const transformProxyResponse = require('node-http-proxy-json')
 
 const stripCookieSecure = cookie => {
     return cookie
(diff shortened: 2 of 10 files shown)


adapter/src/components/LoginModal.js
```diff
@@ -11,10 +11,11 @@ import i18n from '../locales'
 import { post } from '../utils/api'
 
 const staticUrl = process.env.REACT_APP_DHIS2_BASE_URL
+const defaultUrl = process.env.REACT_APP_DHIS2_DEFAULT_BASE_URL
```
mediremi, line 14: We can also use `DHIS2_DEFAULT_BASE_URL` for Netlify PR previews.

cli/src/lib/proxy.test.js
```diff
@@ -0,0 +1,45 @@
...
+            })
+        ).toBe(`${baseUrl}/login.action`)
+
+        expect(
+            rewriteLocation({
+                location: 'https://play.dhis2.org/dev/page?param=value',
+                target: 'https://play.dhis2.org/dev',
+                baseUrl,
+            })
+        ).toBe(`${baseUrl}/page?param=value`)
```
mediremi, line 21: Are there any other edge cases that should be tested here?

varl:
Sorry, @mediremi. I was going over and deleting some temporary comments I put in during my review and accidentally deleted the comment I was responding to (yours). That wasn't intentional. :|

adapter/src/components/LoginModal.js
```diff
@@ -11,10 +11,11 @@ import i18n from '../locales'
...
 
 export const LoginModal = () => {
     const [server, setServer] = useState(
-        staticUrl || window.localStorage.DHIS2_BASE_URL || ''
+        staticUrl || defaultUrl || window.localStorage.DHIS2_BASE_URL || ''
```
varl, line 18: I accidentally removed a comment by @mediremi here about that the `DHIS2_DEFAULT_BASE_URL` can be used for Netlify PR previews.

cli/src/lib/proxy.js
```diff
@@ -0,0 +1,127 @@
...
+    if (
+        parsedLocation.host === parsedTarget.host &&
+        parsedLocation.pathname.startsWith(parsedTarget.pathname)
+    ) {
+        return url.format({
+            ...parsedBaseUrl,
+            pathname: parsedLocation.pathname.replace(
+                parsedTarget.pathname,
+                ''
+            ),
```
amcgee, line 28: Won't this have issues if i.e. the target pathname is `/dev` and there's a `/dev` somewhere OTHER than at the start of the string?  I.e. we should probably replace `/^${pathname}/` instead, so that only the first match is replaced

cli/src/lib/proxy.test.js
amcgee: Maybe cases where the target has a non-80 port?  And/or if the scheme is different (http vs https)

docs/troubleshooting.md
```diff
@@ -8,8 +8,11 @@ A DHIS2 application needs to talk to a [DHIS2 server](https://www.dhis2.org/down
 
 You should specify the fully-qualified base URL of a running DHIS2 server (including the `http://` or `https://` protocol) in the **Server** field of the login dialog.
 
-Make sure that the the URL of your application (`localhost:3000` in this case) is included in the DHIS2 server's CORS whitelist (System Settings -> Access). Also ensure that the server
-you specify doesn't have any domain-restricting proxy settings (such as the `SameSite=Lax`).
+Make sure that the the URL of your application (`localhost:3000` in this case)
+is included in the DHIS2 server's CORS whitelist (System Settings -> Access). If
+your server or browser has domain-restricting settings (such as the
```
amcgee, line 13: This is actually the case in modern Chrome even if the cookies haven't been explicitly set to restrict domains, since they are now SameSite by default (which is also the DHIS2 default)

amcgee (changes requested):
Minor suggestions, this is really awesome @mediremi 🎉

adapter/src/components/LoginModal.js
amcgee: Would be interesting to support a dropdown list of possible backend targets (through the same proxy) here - particularly useful for netlify deploys

cli/src/lib/proxy.test.js
amcgee: Also where the target pathname occurs twice in the location pathname, as mentioned above

cli/src/lib/proxy.js
mediremi: When a string is passed to `String.prototype.replace`, only the first match is replaced (e.g. `'aaa'.replace('a', 'b') === 'baa'`). For it to replace multiple matches, a regex with the `g` flag must be passed (`'aaa'.replace(/a/g, 'b') === 'bbb'`).

adapter/src/components/LoginModal.js
mediremi: We could rename it to `DHIS2_DEFAULT_BASE_URLS` and make it a comma-delimited string of values? If multiple values are passed we render a dropdown otherwise we render an input field?

varl: Are the `LoginModal` changes required for the proxy to work or can we defer adding `REACT_APP_DHIS2_DEFAULT_BASE_URL` and `defaultUrl` until later ?

The Netlify PR deploys do not use the `start` command, they are static builds anyway, so the proxy won't help there right now.

varl: Same with this: https://github.com/dhis2/app-platform/pull/635/files#diff-bcd4b64d0aa7ede2599b25756aa2cb9c9f0e8000857f033edb963b85eb0f54e9R47 

Is that required for the proxy, or can it be deferred until we work on the Netlify deploy feature ?

varl (commented):
Seems to me that the `DHIS2_DEFAULT_BASE_URL` can be omitted from this PR and introduced as a separate feature once we decide how to approach Netlify PR deploys with the proxy.

If that's possible, I'd like to omit it and then we can go ahead and merge this.

adapter/src/components/LoginModal.js
mediremi: It's not required, it's just there to improve the dev UX by prefilling the login modal's server field to the proxy's URL.

varl (approved):

PR #644, opened 2021-09-08
fix(pwa): skip SW update prompt for dev and test env

KaiVandivier:
Showing the 'Update service worker' alert can slow down development somewhat, and can be somewhat problematic during testing because it appears in cypress tests.

This change eschews the prompt to skip waiting to activate newly-installed service workers in development and testing environments, which should increase development quality of life with a minor UX cost that should be negligible during development.

The existing behavior of prompting users to update the service worker is maintained in production environments

https://jira.dhis2.org/browse/TECH-678

### DIFF
--- a/docs/pwa/pwa.md
+++ b/docs/pwa/pwa.md
@@ -53,6 +53,12 @@ When “recording mode” is used and a “recording session” is triggered, th
 
 As a measure to protect against saving faulty recordings, once the recording timeout has elapsed and the recording is moving to finish, the service worker will send a message to the client to confirm that this recording should be completed and saved. The client needs to send back a confirmation message within a 10 second timeout, or the recording will be scrapped. If the service worker receives confirmation, it will save the cached request/response pairs for that section and write some metadata about the recorded section in an IndexedDB.
 
+### Updating the service worker
+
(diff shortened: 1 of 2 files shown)


mediremi (approved):

mediremi:
I wonder if we could use cypress to test the prompt :thinking:

KaiVandivier:
I'm actually not sure about this - an expected reload could (and very well might) cause cypress tests to fail, which it tends to do when I test this in the dashboard app (cc @jenniferarnesen)

I'm now thinking about refactoring the dev-mode service worker: currently a new one is created each time the dev server starts, because a precache manifest is still injected despite not performing any precaching in development; it might be possible instead to use the same service worker repeatedly in dev-mode in apps because the SW network-handling code will be the same for each version of `@dhis2/cli-app-scripts`. I'm not sure if the problem can be completely avoided though because an update is still required with a new version of `cli-app-scripts`, so the integration tests might need to expect and handle a reload at some point either way...

varl:
Marking this as draft to prevent unintentional merges.

KaiVandivier:
Somewhat addressed by https://dhis2.atlassian.net/browse/LIBS-344

PR #652, opened 2021-09-12
feat(adapter): clear sensitive caches on user-change

KaiVandivier:
(WIP - awaiting app runtime function) 

Part of a three-part solution to clearing cached data that might be sensitive between users; see [this issue](https://jira.dhis2.org/browse/DHIS2-11759), this [header bar PR](https://github.com/dhis2/ui/pull/740), and this [runtime PR](https://github.com/dhis2/app-runtime/pull/1002).

### DIFF
--- a/adapter/src/utils/useVerifyLatestUser.js
+++ b/adapter/src/utils/useVerifyLatestUser.js
@@ -1,0 +1,47 @@
+import {
+    useConfig,
+    useDataQuery,
+    clearSensitiveCaches,
+} from '@dhis2/app-runtime'
+import { useState } from 'react'
+
+const USER_QUERY = {
+    user: {
+        resource: 'me',
+        params: { fields: ['id'] },
+    },
+}
+
+const LATEST_USER_KEY = 'dhis2.latestUser'
+
+/**
+ * This hook is used to clear sensitive caches if a user other than the one
+ * that cached that data logs in
+ * @returns {Object} - { loading: boolean }
+ */
+export function useVerifyLatestUser() {
+    const { pwaEnabled } = useConfig()
+    const [finished, setFinished] = useState(false)
+    const { loading, error } = useDataQuery(USER_QUERY, {
+        onComplete: async data => {
+            const latestUserId = localStorage.getItem(LATEST_USER_KEY)
+            const currentUserId = data.user.id
+            if (currentUserId !== latestUserId) {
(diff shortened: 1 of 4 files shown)


adapter/src/utils/useVerifyLatestUser.js
```diff
@@ -0,0 +1,49 @@
...
+            const currentUserId = data.user.id
+            if (currentUserId !== latestUserId) {
+                const cachesCleared = await clearSensitiveCaches()
+                localStorage.setItem(LATEST_USER_KEY, currentUserId)
+
+                if (cachesCleared && pwaEnabled) {
+                    // If this is a PWA app, the app-shell cache will need to
+                    // be restored with a page reload
+                    return window.location.reload()
+                }
```
amcgee, line 38: @KaiVandivier I think this works for now, but I don't think it should be necessary... wouldn't this work as-is if we moved the `useVerifyLastUser` usage up one component level, so it completes before the apps shell fetches anything else?

KaiVandivier: Discussed separately, but copied here for others to see: Unfortunately no - if this hook happened above the AppWrapper and below the ServerVersionProvider, the app shell would be missing `api/system/info` in the app shell cache, and would get stuck at the login modal if the app went offline. If the hook happened above ServerVersionProvider, if a user isn't authenticated, then the hook-consuming component wouldn't render the children as-is

KaiVandivier:
Waiting on https://github.com/dhis2/ui/pull/740 to update UI as well in this PR

KaiVandivier:
Ready for re-review, along with 7.x backport: https://github.com/dhis2/app-platform/pull/662

varl:
@KaiVandivier I'm unsure about how to verify this locally.

I tried opening up the PWA example app and logged in with the user `admin`. I recorded and got a success message that things where cached. I verified that items where cached on my client.

I then deleted the authentication cookie to trigger a login without going through the "logout" flow to change the user. I logged in as `system`, and I see cached items with the same timestamps as for the `admin` user.

The logout flow successfully evicts the cache, but not the user-change flow (on expired auth).

Any tips ?

KaiVandivier:
@varl That _is_ the right way to test it, but it works for me 🤔 hmm

varl:
Looks like starting with `--force` refreshed things enough and now it works fine, thanks !

varl (approved):

PR #654, opened 2021-09-14
fix: add crossorgin use-credentials to manifest.json link

amcgee:
As spotted by @mortenoh - 

>  If the manifest requires credentials to fetch, the crossorigin attribute must be set to use-credentials, even if the manifest file is in the same origin as the current page.

([source](https://web.dev/add-manifest/#link-manifest))

Adding `crossorigin="use-credentials"` to the manifest.json link tag prevents the syntax error at character 1 we see when the browser requests the manifest.  This was caused because the manifest request was redirecting to the login page, as the cookie was stripped from that request

### DIFF
--- a/shell/public/index.html
+++ b/shell/public/index.html
@@ -11,7 +11,7 @@
             manifest.json provides metadata used when your web app is installed on a
             user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
         -->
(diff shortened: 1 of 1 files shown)


varl:
Let's get this in before the major break. :+1:

KaiVandivier (approved):

varl (approved):

PR #668, opened 2021-10-02
feat(cli): validate yarn.lock in 'start' and 'build' commands

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-213


cli/src/commands/deduplicate.js
```diff
@@ -0,0 +1,36 @@
...
+        process.exit(1)
+    }
+
+    const yarnLock = fs.readFileSync(paths.yarnLock, 'utf8')
+    const deduped = fixDuplicates(yarnLock)
+    fs.writeFileSync(paths.yarnLock, deduped)
+
+    const duplicates = listDuplicates(deduped)
+    if (duplicates.size > 0) {
+        reporter.error('Failed to deduplicate the following packages:')
```
mediremi, line 20: Failing to deduplicate with `fixDuplicates` may occur if there are multiple major versions of a dependency.

PR #669, opened 2021-10-03
fix(service-worker): change routing to allow redirect to login page

KaiVandivier:
Addresses https://jira.dhis2.org/browse/LIBS-238

Can be tested like this:
1. Build `pwa` package, `adapter` package, then the pwa-app
2. Deploy to a live instance
3. Visit `/api/apps/pwa-app/` and `/api/apps/pwa-app/index.html`; verify app loads
4. Visit `/api/apps/pwa-app/hi` and verify app loads (tests that subroutes are handled by index.html)
5. At any of the any of the above urls, clear the session cookie and reload - verify the app redirects to the login page, and authenticating returns you to the url you left from
6. Visit `/api/apps/pwa-app//manifest.json` to make sure you can visit static files


pwa/src/service-worker/service-worker.js
```diff
@@ -57,45 +57,79 @@ export function setUpServiceWorker() {
...
+            // If this is a URL that starts with /_, skip.
+            if (url.pathname.startsWith('/_')) {
+                return false
+            }
+
+            // If this looks like a URL for a resource, because it contains
+            // a file extension, skip (unless it's index.html)
+            if (
+                fileExtensionRegexp.test(url.pathname) &&
+                !/index\.html$/.test(url.pathname)
```
mediremi, line 91: The (relatively) new `String.prototype.endsWith` would allow making the intention of this line a bit clearer:

```suggestion
                !url.pathname.endsWith("index.html")
```

KaiVandivier: I like that 🙂

pwa/src/service-worker/service-worker.js
```diff
@@ -57,45 +57,79 @@ export function setUpServiceWorker() {
...
-        // even if you decide not to use precaching. See https://cra.link/PWA
-        precacheAndRoute(self.__WB_MANIFEST || [])
+        // even if you decide not to use precaching. See https://cra.link/PWA.
+        // Includes all built assets and index.html
+        const precacheManifest = self.__WB_MANIFEST || []
+
+        // Extract index.html from the manifest to precache, then route
+        // in a custom way
+        const indexHtmlManifestEntry = precacheManifest.find(({ url }) =>
+            url.includes('index.html')
```
mediremi, line 67: Maybe this should be `url.endsWith('index.html')` too since this would match e.g. `index.html.js`

KaiVandivier: Hah yeah, fair point 🙂

mediremi (approved):

PR #672, opened 2021-10-05
fix(pwa): fix SW update UX

KaiVandivier:
Addresses https://jira.dhis2.org/browse/LIBS-237

- An update prompt is shown instead of reloading automatically on first SW installation
- When a user clicks 'update' while there are >1 tabs open of the app, there is an additional confirmation dialog that warns about unsaved data being lost when all tabs reload

Includes a substantial refactor of `OfflineInterface` to become a better, well, offline interface; so update logic can be moved to a React component in the App Adapter that can use things like alerts and DHIS2 UI elements.  Update alert logic can be removed from the App-Runtime components now too, which simplifies this update flow and makes it much easier to reason about.

To do:
- [x] Remove update alert and `init` handling in App Runtime (https://github.com/dhis2/app-runtime/pull/1041)
- [x] Remove `init` here and bump App Runtime dependency in `shell`

Known issues:

- If a user has multiple tabs open, clicks 'Update' on the alert, then cancels the confirmation dialog, the alert closes and the user would have to refresh the page to see the update alert again

To test:
1. Build `pwa` and `adapter` packages, then start PWA app (with `--force`)
2. (Assuming you've loaded the PWA app before) you should see the 'Update' prompt normally when the app loads -- go ahead and update.
3. **One client/first installation:** In dev tools, make sure there is one client open, unregister the service worker, then reload the page. Verify that the page does _not_ automatically reload when new service worker installs, and an update prompt is shown (Side note: 'recording mode' will not work at this point). Verify that clicking 'Update' reloads the page and recording mode now works. (Side note: refreshing normally works to give the SW control of _this_ tab in this circumstance)
4. **Multiple clients/first installation:** This one is weird; bear with me: In dev tools, unregister the SW, refresh the page, unregister _again_, then duplicate this tab. Both tabs should now install an active service worker, but both are _uncontrolled_. In either tab, click 'Update' on the alert. Verify a confirmation dialog pops up mentioning reloading and data loss, and that clicking 'Reload' reloads both tabs.
5. **Multiple clients/SW update:** Stop and restart the development server. When one tab loads, verify that both tabs show an 'Update' alert. Click 'Update' on one. Verify the 'Reload' confirmation dialog shows, and that both tabs reload when you click 'Reload'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [8.2.1](https://github.com/dhis2/app-platform/compare/v8.2.0...v8.2.1) (2021-10-08)
+
+
+### Bug Fixes
+
+* **shell:** wrap app in CustomDataProvider in App test ([#677](https://github.com/dhis2/app-platform/issues/677)) ([3325605](https://github.com/dhis2/app-platform/commit/332560595d58fc49c3a06e0100029a21738a2cd3))
+
 # [8.2.0](https://github.com/dhis2/app-platform/compare/v8.1.2...v8.2.0) (2021-10-07)
 
 
--- a/adapter/package.json
+++ b/adapter/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dhis2/app-adapter",
-    "version": "8.2.0",
+    "version": "8.2.1",
     "repository": {
         "type": "git",
         "url": "https://github.com/amcgee/dhis2-app-platform",
@@ -21,11 +21,11 @@
         "build"
     ],
     "dependencies": {
-        "@dhis2/pwa": "8.2.0",
+        "@dhis2/pwa": "8.2.1",
         "moment": "^2.24.0"
     },
     "devDependencies": {
-        "@dhis2/cli-app-scripts": "8.2.0",
+        "@dhis2/cli-app-scripts": "8.2.1",
         "@testing-library/react": "^12.0.0",
         "enzyme": "^3.11.0",
         "enzyme-adapter-react-16": "^1.15.5",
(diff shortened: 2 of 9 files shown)


pwa/src/offline-interface/offline-interface.js
```diff
@@ -58,44 +59,88 @@ export class OfflineInterface {
...
+            )
+            const onClick = () => {
+                if (clientsCount <= 1) {
+                    // One client - go ahead and reload
+                    onConfirm()
+                } else if (window.confirm(confirmReloadMessage)) {
+                    // Multiple clients - make sure to confirm reload to avoid
+                    // data loss
+                    onConfirm()
+                }
```
KaiVandivier, line 107: `window.confirm` used here because using React UI components here is complicated -- another alert could be used through the app runtime, but I'm also thinking about a way to refactor much of the update logic to live in React land but use the offline interface for communication; for example a `<PWAUpdateManager />` component in the app adapter

pwa/src/offline-interface/offline-interface.js
```diff
@@ -58,44 +59,88 @@ export class OfflineInterface {
...
+
+        // Receives messages from service worker and forwards to event emitter
+        const handleSWMessage = event => {
+            if (!event.data) {
                 return
             }
+            const { type, payload } = event.data
+            this.offlineEvents.emit(type, payload)
+        }
+        navigator.serviceWorker.addEventListener('message', handleSWMessage)
```
KaiVandivier, line 73: This is just moved up from below so the 'update' logic can send and receive messages to get the number of clients open

PR #674, opened 2021-10-07
feat: provide the app's name & version to the <HeaderBar>

Mohammer5:
Closes LIBS-176 ([link](https://jira.dhis2.org/browse/LIBS-176))
Blocked by dhis2/ui#795

* Passes the app version & name directly to the `<HeaderBar>`
* I didn't want to pass this through the app-runtime:
  * It'd be more work & make the whole thing more complicated
  * ~~With the global shell, the headerbar would have "subscribe" to the values in order to notice the change. With a prop, this happens automatically~~ (might be wrong.. here :facepalm:)
  * The "appName" is already a prop, so making "appVersion" a prop as well seems reasonable to me

### DIFF
--- a/adapter/src/index.js
+++ b/adapter/src/index.js
@@ -7,7 +7,14 @@ import { ServerVersionProvider } from './components/ServerVersionProvider'
 
 const offlineInterface = new OfflineInterface()
 
-const AppAdapter = ({ url, apiVersion, appName, appVersion, pwaEnabled, children }) => (
+const AppAdapter = ({
+    url,
+    apiVersion,
+    appName,
+    appVersion,
+    pwaEnabled,
+    children,
+}) => (
     <ErrorBoundary fullscreen onRetry={checkForSWUpdateAndReload}>
         <ServerVersionProvider
             url={url}
--- a/cli/src/lib/shell/index.js
+++ b/cli/src/lib/shell/index.js
@@ -26,7 +26,11 @@ module.exports = ({ config, paths }) => ({
             cmd: 'yarn',
             args: ['run', 'start'],
             cwd: paths.shell,
-            env: getEnv({ ...getAppInfos(config), port, ...getPWAEnvVars(config) }),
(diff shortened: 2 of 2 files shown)


amcgee (commented):
Hmmm... implementation looks good here.  I'm not sure about making it a prop instead of using the @dhis2/app-service-config context to pass it in...  As mentioned, when moving to a global shell we'll have to "subscribe" to the values with context lets us do.  At that time we'll need to remove or deprecate the appName prop from the HeaderBar anyway (and also appVersion)

I think the main reason I'd like to make this available through `useConfig`, though, is that the application version (and maybe even name as well) could likely be useful in other application components... would it be a huge amount of additional effort to add it there?  I'm not opposed to adding this for now, if it's a lot easier... being able to access the resolved application config could be quite useful I think.

Mohammer5:
> the application version (and maybe even name as well) could likely be useful in other application components

I think putting the information into `useConfig` and expecting these values as a prop are not mutually exclusive.

> when moving to a global shell we'll have to "subscribe" to the values which context lets us do. At that time we'll need to remove or deprecate the appName prop from the HeaderBar anyway (and also appVersion)

I don't understand the reasoning here. From what I can tell, the global shell will always know which app and its version is being displayed (as it's the service responsible for replacing the app when navigating to a different one). Changing the prop would also cause the headerbar to rerender.

---

I think this boils down to a more fundamental aspect:
Should our components retrieve data by themselves or should data be provided (from a certain perspective: should they be controlled or uncontrolled)

I think in the end this is quite a subjective topic.. My experience makes me believe that controlled components are more robust/future-proof without having to anticipate the future (I can expand on why I think this way if you want; I guess this is the part where different developers have made different experiences in their career and therefore favor different approaches). With that in mind I guess using the `useConfig` wouldn't cause any harm, but I don't see the necessity to go that way as it's slightly (not a lot) more work with potential drawbacks.

PR #690, opened 2021-11-12
fix(cli-38): catch unhandled rejections

varl:
- fix(cli-38): catch syntax errors and recover
- fix(cli-38): catch all unhandled rejections

Depends on:
- https://github.com/dhis2/cli-helpers-engine/pull/145
- https://github.com/dhis2/cli-helpers-engine/pull/146

### DIFF
--- a/cli/src/commands/build.js
+++ b/cli/src/commands/build.js
@@ -13,6 +13,10 @@ const makeShell = require('../lib/shell')
 const { validatePackage } = require('../lib/validatePackage')
 const { handler: pack } = require('./pack.js')
 
+process.on('unhandledRejection', err => {
+    throw err
+})
+
 const buildModes = ['development', 'production']
 
 const getNodeEnv = () => {
--- a/cli/src/commands/build.js
+++ b/cli/src/commands/build.js
@@ -13,10 +13,6 @@ const makeShell = require('../lib/shell')
 const { validatePackage } = require('../lib/validatePackage')
 const { handler: pack } = require('./pack.js')
 
-process.on('unhandledRejection', err => {
-    throw err
-})
-
 const buildModes = ['development', 'production']
 
 const getNodeEnv = () => {
--- a/cli/src/lib/compiler/compile.js
+++ b/cli/src/lib/compiler/compile.js
@@ -1,6 +1,6 @@
 const path = require('path')
 const babel = require('@babel/core')
-const { reporter } = require('@dhis2/cli-helpers-engine')
+const { reporter, prettyPrint } = require('@dhis2/cli-helpers-engine')
 const chokidar = require('chokidar')
 const fs = require('fs-extra')
 const makeBabelConfig = require('../../../config/makeBabelConfig.js')
@@ -100,11 +100,12 @@ const compile = async ({
                     babelConfig
                 )
                 await fs.writeFile(destination, result.code)
-                console.info(`Compiling ${source}: OK.`)
             } catch (err) {
-                console.error(err)
-                console.info(
-                    `Compiling ${source}: ERROR. Fix the problem and save the file to automatically reload.`
+                reporter.dumpErr(err)
+                reporter.error(
+                    `Failed to compile ${prettyPrint.relativePath(
+                        source
+                    )}. Fix the problem and save the file to automatically reload.`
                 )
             }
         } else {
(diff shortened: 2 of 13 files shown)


cli/src/lib/compiler/compile.js
```diff
@@ -94,8 +94,19 @@ const compile = async ({
...
+                    source,
+                    babelConfig
+                )
+                await fs.writeFile(destination, result.code)
+                console.info(`Compiling ${source}: OK.`)
+            } catch (err) {
+                console.error(err)
+                console.info(
+                    `Compiling ${source}: ERROR. Fix the problem and save the file to automatically reload.`
+                )
```
varl, line 108: The console messages here are not final, but they can be helpful for the user so that they know _when_ the problematic file is OK again.

cli/src/commands/build.js
```diff
@@ -13,6 +13,10 @@ const makeShell = require('../lib/shell')
 const { validatePackage } = require('../lib/validatePackage')
 const { handler: pack } = require('./pack.js')
 
+process.on('unhandledRejection', err => {
```
varl, line 16: We can add a comment explaining why we do this. It would be great if we could introduce this in a single place in the CLI Engine so we don't have to do this for all our CLI tools. Currently looking into it.

cli/src/lib/compiler/compile.js
```diff
@@ -94,8 +94,19 @@ const compile = async ({
     }
     const compileFile = async (source, destination) => {
         if (source.match(extensionPattern)) {
-            const result = await babel.transformFileAsync(source, babelConfig)
-            await fs.writeFile(destination, result.code)
+            try {
```
varl, line 97: Catching here allows us to recover from a syntax error and re-compile when a change is detected, which is nice for the developer.

cli/src/commands/build.js
KaiVandivier: A comment sounds good -- could maybe link to the [comment](https://jira.dhis2.org/browse/CLI-38?focusedCommentId=45563&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-45563) you added to the Jira issue, or even add that info to the PR and link here which I think I like better

cli/src/lib/compiler/compile.js
KaiVandivier: Confirmed this works in Node 16.13.0 👍

cli/src/lib/compiler/compile.js
KaiVandivier: Should these use the `reporter` logger for consistency?

One suggestion: using the the file's relative path from the working directory would keep the logs nicer; `source` seems to be an absolute path which can be quite long.

In a project with lots of files, these logs can be quite verbose at startup of the dev server when all the file paths in the project logged -- not functionally critical, but it would be nice aesthetically if those can be skipped

KaiVandivier:
Thanks for tackling this! 😁 🙏

cli/src/lib/compiler/compile.js
varl: > In a project with lots of files, these logs can be quite verbose at startup of the dev server when all the file paths in the project logged -- not functionally critical, but it would be nice aesthetically if those can be skipped

Agreed. If we remove the "OK" message now, then the user won't get a message that the compilation succeeded, and the last message in the terminal will be an error message, even though the file has recompiled OK and is ready to go. Probably not worth the upfront spam though.

varl: > Should these use the `reporter` logger for consistency?

Yes, definitely. I left them as console logs to make it clear that these are not final, for the above reasons. :)

cli/src/commands/build.js
amcgee: @varl we should be able to do it here I think?  https://github.com/dhis2/cli-helpers-engine/blob/master/lib/makeEntryPoint.js

varl: @amcgee that's right, I've tested that it works for this case. Wanted to test a bit more in dhis2/cli. Will post a new PR in engine and then link it to this one after verification.

varl: Confirmed, works in other places too. Here's the PR:

https://github.com/dhis2/cli-helpers-engine/pull/145

cli/src/lib/compiler/compile.js
varl: Neato !

cli/src/lib/compiler/compile.js
varl: Removed the "OK" message because of the spam.

Changed to a relative path print for `source`.

Changed to use reporter.

HendrikThePendric (approved):
Awesome!

PR #692, opened 2021-11-19
docs(d2-config): add id and version fields to d2-config reference

Birkbjo:


Mohammer5 (approved):

amcgee (approved):
One small suggestion, LGTM!

PR #701, opened 2022-02-17
feat(cli): build plugins

mediremi:
## TODO

- [ ] Implement method to pass props to plugin
  + [ ] Use `initialProps` iframe attribute containing JSON and then plugin wrapper inside iframe can use `JSON.parse(window.frameElement.getAttribute('initialProps'))` to get props
- [ ] `<Plugin pluginProps={...} />` component provided by `app-runtime` that injects `iframe` with correct `src` and `initialProps` attributes
  + [ ] Bump `app-runtime`'s `cli-app-scripts` peer dep
- [ ] Documentation
  + [ ] Setting up a plugin entrypoint
  + [ ] Using `<Plugin />`
- [ ] Fix hot reload
- [ ] Support service workers (need to update `webpack` config with [these lines](https://github.com/facebook/create-react-app/blob/v4.0.3/packages/react-scripts/config/webpack.config.js#L712-L724))
  + [ ] Will need to ensure scope of service workers set to `/plugin.html` so don't conflict with app service workers

## Usage
In `d2.config.js`, set `type` to `"app"` and provide a `plugin` entrypoint:

```js
{
  type: 'app',
  entryPoints: {
    plugin: './src/plugin.js'
  }
}
```

### Development
`yarn start` and visit `http://localhost:3001/plugin.html`

### Bundle
`yarn build`


cli/src/lib/plugin/start.js
```diff
@@ -0,0 +1,42 @@
+// Based on CRA start script
```
mediremi, line 1: See https://github.com/facebook/create-react-app/blob/v4.0.3/packages/react-scripts/scripts/start.js

cli/src/lib/plugin/webpack.config.js
```diff
@@ -0,0 +1,255 @@
+// Based on CRA Webpack config
```
mediremi, line 1: See https://github.com/facebook/create-react-app/blob/v4.0.3/packages/react-scripts/config/webpack.config.js

The Webpack config from react-scripts v4.0.3 was made for Webpack 4, but this PR uses Webpack 5, and so could not be used directly.

adapter/src/components/AppWrapper.js
```diff
@@ -20,7 +20,7 @@ export const AppWrapper = ({ appName, children, offlineInterface }) => {
     return (
         <div className="app-shell-adapter">
             <style jsx>{styles}</style>
-            <HeaderBar appName={appName} />
+            {!plugin && <HeaderBar appName={appName} />}
```
mediremi, line 23: Don't render the headerbar for plugins

cli/src/lib/plugin/webpack.config.js
mediremi: The main thing missing from this config is support for service workers. This is how react-scripts does it: https://github.com/facebook/create-react-app/blob/v4.0.3/packages/react-scripts/config/webpack.config.js#L712-L724

cli/src/lib/plugin/webpack.config.js
```diff
@@ -0,0 +1,255 @@
...
+            rules: [
+                {
+                    oneOf: [
+                        {
+                            test: /\.(js|mjs|jsx|ts|tsx)$/,
+                            include: paths.shellSrc,
+                            use: {
+                                loader: require.resolve('babel-loader'),
+                                options: {
+                                    ...makeBabelConfig({ mode: webpackEnv }),
```
mediremi, line 181: This webpack config could also be used for apps, eliminating the need for a separate babel transformation pass.

amcgee: This would also allow for code-splitting between app and plugin, which would be advantageous

KaiVandivier:
Merging to alpha to create a release for testing

PR #711, opened 2022-03-30
refactor: use creact-react-app/react-scripts instead of .d2 folder magic

Mohammer5:
Hey guys, I was working on this sort of in secret hehe.. I felt somewhat annoyed by the `.d2` folder and the issues that can happen with it as well as the double file watching (ours and the of of cra inside .d2), the discrepancy between `d2-app-scripts test` and `react-scripts test` and some minor things.

So I was wondering how we could do things differently if we used create-react-app directly. I haven't changed any of the public api of `d2-app-scripts`. Everything still works the same (with one difference: the `src/index.js` and the app-shell; but more on that a bit later).

# Here are the reasons why I think this is a good change

* No more double file watching
* Relying directly on create-react-app's tools without having to implement certain things ourselves (like file watching, test command, building the app)
* The `d2-app-scripts test` command is more or less an alias for `react-scripts test` so we'll get all the features out of the box
* Some code from `cli` was removed, reducing the complexity in that part of the code (note: the reason why there are more lines added than deleted is mostly due to the `cra-template-dhis2-app` being added)
* We have direct control over the `public` folder (e.g. adding scripts like the CKEditor in the maintenance app)

# Here are some challenges that have to be considered

* This change introduces a `src/index.js`
  * IMO this is not a bad change, but actually a good one
  * We'd have to think about how we can let users transition from the old style to new one (should the `init` script be able to update as well?)
* The `template.json` of the cra template (which is used to define what's in the final `package.json` of the app) doesn't copy all values ([see here](https://github.com/dhis2/app-platform/blob/edb6cdfc5b6f220231139074b815b9711064286f/cli/src/lib/init/init-app.js#L6:L25))

# Here are some fundamental changes

### The app-shell is now just a component

* The `<Shell />` wraps the app in the `src/index.js`
* The two style rules have been moved to the `<app-root>/public/index.html`

### Nudging instead of IoC (with the shell)

In theory the user could mess with the `index.js`, I've added a prominently visible comment that - in essence - says something like "DO NOT TOUCH THIS FILE!".

The reason why I use the term "nudging" is:
* Users could modify things that are now closed right now
* It is fairly hard to do that though as you really have to be a dhis2 expert and replicate the shell
* This will nudge users into going the way we're "prescribing"
* This is very close to our philosophy of "Trivial things should be easy, complicated things should be possible"

We could use `react-rewire` to hide the index.js but my reasons for not going that way are:
* we'd have to mess with cra's internal tooling, we'd introduce a point of failure if they change that
* it removes the possibility of doing things in the `index.js` if that's ever needed (it shouldn't be, hence the comment I added)
* it'd increase the complexity of our codebase

# Open ToDos

* [ ] Add pwa back (I haven't done that yet because I want to understand the code first)
* [ ] Figure out the differences between our approach and the [cra pwa approach](https://create-react-app.dev/docs/making-a-progressive-web-app/)
* [ ] Add `spawn` to `@dhis2/cli-helpers-engine`
* [ ] Address all remaining `@TODO` comments


amcgee:
Alternative approaches, some of which I mentioned today:

1. Remove the `create-react-app` indirection and run our own compile/bundle steps directly on the source files, as we currently do for libraries, so that we don't have to work around the limitations of CRA.  We could even reuse some of the CRA internals (a la ejecting)
2. Build a dynamic module instead and run the shell unmodified, then dynamically load the bundled module
3. Use a more flexible intermediary like `nextjs` which provides some of the features we want (including the ability to customize build steps)

An important consideration is security and sandboxing of applications, we need to be able to provide "guarantees" to people installing DHIS2 applications.  This leads to the IoC paradigm we chose, which allows us to control and restrict the problem space of applications (while trying to keep that restriction as flexible as possible).  We also want to make sure we're not pushing complexity from the platform to apps, we should strongly prefer encompassing complexity in the platform rather than proliferating it in hundreds of applications and branches.  The many applications built with our platform

PR #719, opened 2022-06-08
fix(proxy): ignore nulls in transformJsonResponse

Birkbjo:
A bug caused objects that had `property: null` to be converted to `{}`. This was causing the `data-entry-app` to crash, and was a pretty obscure bug to track down. The root of the issue is that `typeof null === 'object'`, and when using `_transform` with a default of `{}`, all `null`-properties were transformed to `{}`.


Mohammer5 (approved):

amcgee:
Good catch, nice investigation @Birkbjo !

PR #728, opened 2022-07-25
chore: beta to stable

ismay:
CI is failing because of the commit checks. Those are failing because of the long descriptions that semantic release generates and appends to the commit description. Not a problem though, won't fail once it's merged to master.


Mohammer5 (approved):
LGTM, example apps are working as expected

amcgee:
> (We should probably also look into disabling commit checks when merging between release branches)
Maybe we could just skip any commits starting with with `chore(release):` as those have changelog bodies...?

PR #740, opened 2022-08-24
chore: merge alpha to master (cli plugin builds)

KaiVandivier:
Intended to show history of CLI build commits and view diffs

Jira Issue: https://dhis2.atlassian.net/browse/TECH-474


cli/src/lib/parseConfig.js
```diff
@@ -43,6 +43,7 @@ const validateConfig = (config) => {
             )
         }
     })
+    // Todo: make sure apps with plugins have `pwa_enabled: true` in the config
```
amcgee, line 46: What is the consequence of not checking this yet?

cli/src/lib/plugin/start.js
```diff
@@ -0,0 +1,47 @@
...
+process.env.NODE_ENV = 'development'
+
+const getPublicUrlOrPath = require('react-dev-utils/getPublicUrlOrPath')
+const webpack = require('webpack')
+const WebpackDevServer = require('webpack-dev-server')
+const webpackConfigFactory = require('./webpack.config')
+
+module.exports = async ({ port, config, paths }) => {
+    const webpackConfig = webpackConfigFactory({
+        // todo: change to development, but this creates a compilation error
```
amcgee, line 14: Can you explain why that would cause a compilation error?  What are the downsides of building for production here?

cli/src/lib/plugin/webpack.config.js
```diff
@@ -0,0 +1,316 @@
+// Based on CRA Webpack config
```
amcgee, line 1: Were there any significant changes you had to make to the CRA config?

amcgee: Also, should this maybe live in the /config directory with our other config files?

cli/src/lib/plugin/webpack.config.js
```diff
@@ -0,0 +1,316 @@
...
+                    swSrc: paths.shellBuildServiceWorker,
+                    injectionPoint: 'self.__WB_PLUGIN_MANIFEST',
+                    // Skip compiling the SW, which happens in the app build step
+                    compileSrc: false,
+                    dontCacheBustURLsMatching: /\.[0-9a-f]{8}\./,
+                    exclude: [
+                        /\.map$/,
+                        /asset-manifest\.json$/,
+                        /LICENSE/,
+                        // TODO dhis2: locales are weird in the plugin build -
```
amcgee, line 206: Any idea why?

amcgee (commented):
Generally looks good!  Left a few comments/questions

cli/src/lib/plugin/start.js
KaiVandivier: It throws the error in Figure 1, which seems to be a conflict with the dev server for both the app and the plugin writing files to the same place? I didn't look into it much because it works in `'production'` mode -- the downsides are quality of life things:
1. It’s harder to debug errors in the plugin because the files are minified and the source maps aren’t working (Figure 2)
2. There’s more webpack chatter in the terminal as it rebuilds for every code change (Figure 3)
3. I’m guessing the plugin has a bit slower start and hot reload times due to the extra compilation steps

Figure 1: `'development'` mode error
plugin-start-fig-1

Figure 2: difficult debugging
plugin-start-fig-2

Figure 3: webpack chatter
plugin-start-fig-3

cli/src/lib/parseConfig.js
KaiVandivier: Ah this is an obsolete comment -- thanks for catching it! Plugins don't need to enable PWA, but they do if they will be used in PWA parent apps

cli/src/lib/plugin/webpack.config.js
KaiVandivier: I think I wrote this because all the generated locales files are duplicated (but slightly different); one batch in a `moment-locales` dir and another in `plugin-moment-locales`

I think we talked about not precaching locales files too, conveniently

It's been a while since I looked at this, so I forget if there was anything else 🤔

cli/src/lib/plugin/webpack.config.js
KaiVandivier: Two main changes:
1. Handling of fonts and other non-js files
2. Modifying the 'InjectManifest' plugin

I tried to mark these with `// dhis2:` comments; in [this PR to incorporate the feedback](https://github.com/dhis2/app-platform/pull/773) I added one to line 279 🙂

amcgee (approved):

KaiVandivier:
Some follow-up issues to this PR are created here: https://dhis2.atlassian.net/browse/LIBS-390

PR #745, opened 2022-09-09
feat: include plugin launch path plugin.html in built manifests [LIBS-346]

amcgee:
Fixes [LIBS-346](https://dhis2.atlassian.net/browse/LIBS-346)

Generate `plugin_launch_path` in `manifest.webapp` which will be served by the backend in the `api/apps` API

Also includes a modified `entrypoints` map in the `d2.config.json` output:

```json
"entrypoints": {
    "app": "index.html"
    "plugin": "plugin.html"
}
```

### DIFF
--- a/cli/src/lib/generateManifests.js
+++ b/cli/src/lib/generateManifests.js
@@ -108,7 +108,8 @@ module.exports = (paths, config, publicUrl) => {
         version: config.version,
         core_app: config.coreApp,
 
-        launch_path: 'index.html',
+        launch_path: paths.launchPath,
+        plugin_lauch_path: paths.pluginLaunchPath,
         default_locale: 'en',
         activities: {
             dhis: {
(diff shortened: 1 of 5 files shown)


cli/src/lib/generateManifests.js
```diff
@@ -108,7 +108,8 @@ module.exports = (paths, config, publicUrl) => {
         version: config.version,
         core_app: config.coreApp,
 
-        launch_path: 'index.html',
+        launch_path: paths.launchPath,
+        plugin_lauch_path: paths.pluginLaunchPath,
```
edoardo, line 112:
```suggestion
        plugin_launch_path: paths.pluginLaunchPath,
```

KaiVandivier: ah good catch

amcgee: 😬 yikes thanks @edoardo !

PR #748, opened 2022-09-22
feat: headerbar update notification [LIBS-344]

amcgee:
Implements [LIBS-176](https://dhis2.atlassian.net/browse/LIBS-176)
Implements [LIBS-344](https://dhis2.atlassian.net/browse/LIBS-344)

~Requires https://github.com/dhis2/app-runtime/pull/1252~
~Requires https://github.com/dhis2/ui/pull/1153~ released as `beta`
~Requires https://github.com/dhis2/ui/pull/1156 (merge `ui` beta to master)~

## Description

This refactors the PWA update notification logic, moving update notification UI to the Profile menu.  It also adds the app name, app version, and server version strings to the `useConfig` context.


KaiVandivier (changes requested):
Looks really good!! I noticed one bug: when hard-reloading in the one-client condition, the page got stuck at the PWA Loading Boundary (white screen) until reloading again. I found that removing [these lines](https://github.com/dhis2/app-platform/blob/4a5d59b45348e5058b07d5ff33014ab60e45630b/pwa/src/lib/registration.js#L190-L197) fixed it -- they were the old way of handling the hard-reload

After that's fixed, this is functionally good to go -- I'll post a few ideas for follow-ups and suggestions for making one or two things easier to understand for future maintainers too, but they're not at all necessary for merging 🙂

KaiVandivier (approved):

PR #753, opened 2022-10-03
chore: fix cli-style useEffect exhaustive deps warnings

amcgee:
Avoid cli-style warnings and clean up useEffect calls to make logic a bit cleaner

### DIFF
--- a/adapter/src/components/Alerts.js
+++ b/adapter/src/components/Alerts.js
@@ -25,8 +25,8 @@ const Alerts = () => {
         )
 
     useEffect(() => {
-        if (alertManagerAlerts.length > 0 || alertStackAlerts.length > 0) {
-            setAlertStackAlerts(
+        if (alertManagerAlerts.length > 0) {
+            setAlertStackAlerts(alertStackAlerts =>
                 mergeAlertStackAlerts(alertStackAlerts, alertManagerAlerts)
             )
         }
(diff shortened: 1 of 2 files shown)


adapter/src/components/Alerts.js
```diff
@@ -25,8 +25,8 @@ const Alerts = () => {
         )
 
     useEffect(() => {
-        if (alertManagerAlerts.length > 0 || alertStackAlerts.length > 0) {
-            setAlertStackAlerts(
+        if (alertManagerAlerts.length > 0) {
+            setAlertStackAlerts((alertStackAlerts) =>
```
Mohammer5, line 29: Very minor nitpick: This variable has the same name as the variable returned by `useState` in the outer scope. Maybe `currentAlertStackAlert` (or another name that ensures that the reader immediately recognizes that this is a different variable)?

Mohammer5 (approved):

adapter/src/components/Alerts.js
amcgee: Yeah that's a nice readability improvement 👍

adapter/src/components/Alerts.js
```diff
@@ -25,8 +25,8 @@ const Alerts = () => {
...
                 mergeAlertStackAlerts(alertStackAlerts, alertManagerAlerts)
             )
```
amcgee, line 31:
```suggestion
            setAlertStackAlerts((currentAlertStackAlerts) =>
                mergeAlertStackAlerts(currentAlertStackAlerts, alertManagerAlerts)
            )
```

PR #757, opened 2022-10-17
fix: check authorities in app adapter [LIBS-370]

KaiVandivier:
Address [LIBS-370](https://dhis2.atlassian.net/browse/LIBS-370)

Includes the fixes for [LIBS-356](https://dhis2.atlassian.net/browse/LIBS-356) and [LIBS-357](https://dhis2.atlassian.net/browse/LIBS-357) (which were split off from here and merged in separate PRs)

Checks authorization on the client side by using an AuthBoundary component to check correct authorities and block the app if necessary. This is needed because some deployment environments don't return a 403 'Forbidden' response (like standalone deployments and Netlify, it seems)

Todo:
- [x] Get correct app name (and maybe format) for auth checks from CLI and add to app env
- [x] Add UI for "Forbidden" screen instead of just hitting the error boundary
- [x] Clean up useVerifyLatestUser

UI Draft, updated:
Screen Shot 2022-10-18 at 9 22 15 AM

[LIBS-370]: https://dhis2.atlassian.net/browse/LIBS-370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIBS-356]: https://dhis2.atlassian.net/browse/LIBS-356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIBS-357]: https://dhis2.atlassian.net/browse/LIBS-357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adapter/src/components/AppWrapper.js
```diff
@@ -1,32 +1,33 @@
...
 import { Alerts } from './Alerts.js'
+import { AuthBoundary } from './AuthBoundary.js'
 import { ConnectedHeaderBar } from './ConnectedHeaderBar.js'
 import { ErrorBoundary } from './ErrorBoundary.js'
 import { LoadingMask } from './LoadingMask.js'
 import { styles } from './styles/AppWrapper.style.js'
 
 export const AppWrapper = ({ children }) => {
     const { loading: localeLoading } = useCurrentUserLocale()
-    const { loading: latestUserLoading } = useVerifyLatestUser()
```
KaiVandivier, line 13: Moved to `<AuthBoundary>`

adapter/src/components/AuthBoundary.js
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+}
+
+const isAppAvailable = (authorities) => {
+    // Skip check on dev
+    // TODO: should we check on dev environments too?
+    if (!IS_PRODUCTION_ENV) {
+        return true
+    }
```
KaiVandivier, line 44: Curious what others think -- most app devs probably have admin privileges, but if they don't have app management authority, it's possible that someone might end up working on an a new app which they don't have authorities for.

Someone might be able to bypass their authorities limitations by running an app locally though?...

Currently I guess we don't do any checks like this

cli/src/lib/formatAppAuthName.js
```diff
@@ -0,0 +1,30 @@
...
+ * Ex: coreApp && name = 'data-visualizer': authName = 'M_dhis-web-data-visualizer'
+ * Ex: name = 'pwa-example': authName = 'M_pwaexample'
+ * Ex: name = 'BNA Action Tracker': authName = 'M_BNA_Action_Tracker'
+ */
+const formatAppAuthName = (config) => {
+    if (config.coreApp) {
+        // TODO: Verify this formatting - are there any transformations,
+        // or do we trust that it's lower-case and hyphenated?
+        return APP_AUTH_PREFIX + DHIS_WEB + config.name
+    }
```
KaiVandivier, line 17: Curious about this formatting

pwa/src/service-worker/service-worker.js
```diff
@@ -143,7 +144,8 @@ export function setUpServiceWorker() {
         ({ url }) =>
             PRODUCTION_ENV &&
             urlMeetsAppShellCachingCriteria(url) &&
-            fileExtensionRegexp.test(url.pathname),
+            fileExtensionRegexp.test(url.pathname) &&
+            !/\.(json|action)/.test(url.pathname), // don't SWR these
```
KaiVandivier, line 148: LIBS-357

pwa/src/service-worker/service-worker.js
```diff
@@ -99,17 +99,18 @@ export function setUpServiceWorker() {
...
                 .then((response) => {
-                    if (response.type === 'opaqueredirect') {
-                        // It's sending a redirect to the login page. Return
-                        // that to the client
+                    if (response.type === 'opaqueredirect' || !response.ok) {
+                        // It's sending a redirect to the login page,
+                        // or an 'unauthorized'/'forbidden' response.
+                        // Return that to the client
                         return response
                     }
```
KaiVandivier, line 107: Pass 403 responses to the client in production, instead of returning index.html

adapter/src/components/AuthBoundary.js
```diff
@@ -0,0 +1,112 @@
...
+import React, { useState } from 'react'
+import i18n from '../locales'
+import { LoadingMask } from './LoadingMask'
+import styles from './styles/ErrorBoundary.style.js'
+
+const LATEST_USER_KEY = 'dhis2.latestUser'
+const IS_PRODUCTION_ENV = process.env.NODE_ENV === 'production'
+
+const APP_MANAGER_AUTHORITY = 'M_dhis-web-maintenance-appmanager'
+const REQUIRED_APP_AUTHORITY = process.env.REACT_APP_DHIS2_APP_AUTH_NAME
```
KaiVandivier, line 17: Added in CLI

adapter/src/components/AuthBoundary.js
KaiVandivier: (You can make testing easier by commenting out the `return true` to run it on dev)

KaiVandivier:
You can test the boundary by logging in as a guest user (user guest, pass Guest123), which shouldn't have access to all apps -- also see the 'production env' test above that you can comment out to test it in a dev environment locally

adapter/src/components/AuthBoundary.js
HendrikThePendric: I initially wasn't sold on the idea of always returning true in a development environment, but think I'm coming around to it.
|         | Full reload     | From SW |
|--------------|-----------|------------|
| Production | Server renders no-access page      | App-shell renders no-access view        |
| Development      | Dev can access app  | Dev can access app       |

adapter/src/components/AuthBoundary.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+    // Check for three possible authorities
+    return authorities.some((authority) =>
+        ['ALL', APP_MANAGER_AUTHORITY, REQUIRED_APP_AUTHORITY].includes(
+            authority
+        )
+    )
+}
+
+const ForbiddenScreen = ({ appName, baseUrl }) => {
```
HendrikThePendric, line 53: Maybe consider renaming. Perhaps something like `AccessErrorOverlay` / `ForbiddenOverlay` would be more accurate. I guess my main qualm is with the word `Screen`.

HendrikThePendric (commented):
Looks very Kai. Thanks for noticing this had slipped through the cracks and turning this around so quickly. I've left a few comments.

adapter/src/components/AuthBoundary.js
```diff
@@ -0,0 +1,112 @@
...
+    return (
+        <div className="mask fullscreen">
+            <style jsx>{styles}</style>
```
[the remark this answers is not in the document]
KaiVandivier: Yeah it does currently cover the header bar -- I was also considering what you suggested though, and I think it's better. I was wondering if header bar contents might be sensitive in some way, but I don't think so

I'll refactor this to wrap just the app

KaiVandivier:
@amcgee I noticed that -- I refactored the app wrapper in my latest commit to work well with the changes that you have here: https://github.com/dhis2/app-platform/pull/754/files#diff-58fcbeed2ec3103f3d186cb172bf2da570e316e008fb614240d6aa0f72df7574

cli/src/lib/formatAppAuthName.js
amcgee: We do need to do transformation here, as the "name" in the d2 config is NOT necessarily what's used for the authority (or URL name) - see [here](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java#L494-L499).  We already replicate this logic in the platform [here](https://github.com/dhis2/app-platform/blob/master/cli/src/lib/constructAppUrl.js)

amcgee: Ah I see we're doing that below - but we should also do it for core apps.

adapter/src/components/AppWrapper.js
KaiVandivier: ...unmoved and reimplemented here

adapter/src/components/AuthBoundary.js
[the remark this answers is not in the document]
KaiVandivier: New UI:
Screen Shot 2022-10-17 at 6 21 22 PM

pwa/src/service-worker/service-worker.js
amcgee: There are other extensions we should probably except here - wouldn't it be easier to use `/api` and other base paths instead of trying to identify extensions?

amcgee (changes requested):
Would be good to not duplicate logic between this and the session expiration PR

cli/src/lib/formatAppAuthName.js
KaiVandivier: Yeah, notice that the formatting is different between those two things you linked, namely around hyphens -- core apps seem to be getting a different treatment?

pwa/src/service-worker/service-worker.js
KaiVandivier: That could work; do you have a sense of what those rules might be?

I'm wondering if there might be requests to external URLs for `json`, for example

adapter/src/components/AuthBoundary.js
```diff
@@ -0,0 +1,50 @@
...
+        children
+    ) : (
+        <CenteredContent>
+            <NoticeBox error title={i18n.t('Forbidden')}>
+                {i18n.t(
+                    "You don't have access to the {{appName}} app. Contact your system administrator if this seems to be an error.",
+                    { appName }
+                )}
+            </NoticeBox>
+        </CenteredContent>
```
cooper-joe, line 42: I suggest changing this message slightly to avoid the `Forbidden` noticebox title.

Title: You don't have access to the {{appName}} app
Body text: Contact your system administrator for assistance with app access.

KaiVandivier: Haha right on, maybe "Forbidden" is too harsh

KaiVandivier: Updated text and screenshot 🙂

cli/src/lib/formatAppAuthName.js
KaiVandivier: Added some formatting insurance

KaiVandivier:
Also added handling for server versions < 35, as in https://github.com/dhis2/app-platform/blob/fix/pwa-handle-navigations/cli/src/lib/constructAppUrl.js#L8

KaiVandivier:
Drafted while this is on the back-burner

pwa/src/service-worker/service-worker.js
KaiVandivier: Addressed this with a different approach here: https://github.com/dhis2/app-platform/pull/762

KaiVandivier:
I think concerns have been addressed and this is ready for rereview @amcgee

PR #762, opened 2022-10-21
fix(pwa): file SWR filter & allow navigation 403s [LIBS-356] [LIBS-357]

KaiVandivier:
Pared down version of #757 
Addresses https://dhis2.atlassian.net/browse/TECH-1462  (partially)

* Switches SWR filter to handle just image files
* Passes 403 responses to client to show 'unauthorized' when online

Pairs well with https://github.com/dhis2/app-runtime/pull/1268 🥂

### DIFF
--- a/pwa/src/service-worker/service-worker.js
+++ b/pwa/src/service-worker/service-worker.js
@@ -144,7 +144,7 @@ export function setUpServiceWorker() {
         ({ url }) =>
             PRODUCTION_ENV &&
             urlMeetsAppShellCachingCriteria(url) &&
-            /\.(jpg|gif|png|bmp|tiff|ico)$/.test(url.pathname),
+            /\.(jpg|gif|png|bmp|tiff|ico|woff)$/.test(url.pathname),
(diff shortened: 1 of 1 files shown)


amcgee (approved):
I think this looks good, nice and simple!  I'm approving, but let's discuss super quickly what the implications of changing this are, just so we make sure it's super low-risk for 2.39.0

KaiVandivier:
Summary of in-person discussion:

We'll go with this minimal fix instead of PR #757 (which includes some client-side changes) with the intention of keeping it simple for 2.39.0. This simpler version will address unauthorized users from viewing pwa apps while online, but there will still be some obscure situations where an unauthorized user can see a pwa app while offline.  

The SWR changes will fix some issues that we saw by only handling images; the performance benefit that we want is really only on images anyway. Caching `.action` or `.json` requests could be problematic or sensitive.

https://github.com/dhis2/app-runtime/pull/1268 will also help avoid issues with the SWR strategy showing requests belonging to another user.

PR #779, opened 2023-01-16
fix(plugins): fix file loader behavior [LIBS-399]

KaiVandivier:
Fixes https://dhis2.atlassian.net/browse/LIBS-399

It seems like the `file-loader` is mangling the css file from maplibre-gl (which doesn’t need compiling, strangely) and outputting all kinds of nasty junk that doesn’t work, like:
* `static/media/javascript,__webpack_public_path__=….bin`
* `static/media/svg+xml;base64,ASD:LKFDNAJKSD….bin`
* other nasty `.bin` files

In the maps app, this caused an "Filename too long" error that prevented building the app, and `svg` icons were broken when run in dev mode.

Changing to the modern asset/resource "file loader" seems to fix it with no noticeable side effects.

This configuration was originally what I used when first making plugins, but I seem to remember there was a reason not to use it -- I think it had to do with CSS modules, but this config seems to be running and building fine in the PWA test app, the LL app, and the Maps app, so I don't see any reason not to use it now 🤷‍♂️

### DIFF
--- a/cli/config/plugin.webpack.config.js
+++ b/cli/config/plugin.webpack.config.js
@@ -105,9 +105,6 @@ module.exports = ({ env: webpackEnv, config, paths }) => {
             chunkFilename: isProduction
                 ? 'static/js/plugin-[name].[contenthash:8].chunk.js'
                 : 'static/js/plugin-[name].chunk.js',
-            // ! dhis2: this at least gets fonts to match the CRA build,
-            // but is re-outputting them
-            assetModuleFilename: 'static/media/[name].[hash][ext]',
             // TODO: investigate dev source maps here (devtoolModuleFilenameTemplate)
         },
         optimization: {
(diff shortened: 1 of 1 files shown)


KaiVandivier:
@jenniferarnesen has confirmed that this fix works in the maps app 👍

cli/config/plugin.webpack.config.js
```diff
@@ -275,33 +272,25 @@ module.exports = ({ env: webpackEnv, config, paths }) => {
...
                         // that fall through the other loaders.
                         {
-                            loader: require.resolve('file-loader'),
                             // Exclude `js` files to keep "css" loader working as it injects
                             // its runtime that would otherwise be processed through "file" loader.
                             // Also exclude `html` and `json` extensions so they get processed
-                            // by webpacks internal loaders.
+                            // by webpack's internal loaders.
                             exclude: [
+                                /^$/,
```
HendrikThePendric, line 286: I believe this newly added regex only matches an empty string. I wonder if this is actually useful, or perhaps a remnant of some early stage experimentation.....
Screenshot 2023-02-16 at 15 05 24



Could you explain why this is needed here?

HendrikThePendric (commented):
I left a comment because one part was unclear to me. Perhaps I am simply failing to see what's going on. So I've just left a comment, instead of approving or requesting changes....

HendrikThePendric (approved):
I'm approving after a brief discussion on Slack. @KaiVandivier and I both do not fully understand why it might be useful to exclude empty strings in the file loader. But since CRA also include this one, we might as well do the same.

PR #785, opened 2023-02-14
fix: bump header bar in app shell [DHIS2-8651]

tomzemp:
Bumps ui version in app shell to pick up new header bar version (bumped to latest version to pick up correct build of multi-calendar-dates package)


KaiVandivier (approved):
Looks good 🙂 you could even bump it again to 8.11.0 if you want, but I don't think there's anything that would affect the headerbar

PR #786, opened 2023-02-20
feat: implement plugin wrapper (receiver in plugin)

tomzemp:
This PR is conceived in conjunction with [PR in app-runtime](https://github.com/dhis2/app-runtime/pull/1332) to provide plugin wrappers.

The main change in this code is adding PluginLoader in the app shell which sets up communication with the parent window. For the communication to work, PluginSender needs to be used in the app, otherwise, the communication will fail (the plugin would still load, so you can still build a plugin with this updated app shell and not use PluginSender if you do not want this communication channel, but do want a separate plugin build)

There is some additional updates in app/adapter for hoisting plugin alerts and errors.

**Some notes**

- Error boundaries are to be designed by Joe. For now, I have just modified the existing Error Boundary to make it work with a plugin
- I'm not really sure what we want to do with checks for required props. Previously when the wrapper within the plugin was in app-runtime, I set it up to take a requiredProps parameter. Now I've set it up to read requiredProps from d2.config.js (which I do not like from a practical perspective as it is now putting these into an environment variable (string) and reparsing). I feel like, it would be better to just let people check for required props in the plugin rather than trying to incorporate that as a default process into the build logic of the plugin.  If we want this to be easier, we could create a custom hook in app-runtime/services/plugin that lets you pass your required props and then will throw an error if they are missing?

### DIFF
--- a/shell/src/App.js
+++ b/shell/src/App.js
@@ -1,10 +1,25 @@
 import AppAdapter from '@dhis2/app-adapter'
 import { Layer, layers, CenteredContent, CircularLoader } from '@dhis2/ui'
+import PropTypes from 'prop-types'
 import React from 'react'
+import { PluginLoader } from './PluginLoader'
+import { PluginOuterErrorBoundary } from './PluginOuterErrorBoundary'
 
 const D2App = React.lazy(() =>
     import(/*webpackChunkName: 'app'*/ './D2App/app')
 ) // Automatic bundle splitting!
+
+const parseRequiredProps = (propsEnvVariable) => {
+    if (!propsEnvVariable || propsEnvVariable === '') {
+        return []
+    }
+    return propsEnvVariable.split(',')
+}
+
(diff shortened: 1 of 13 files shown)


shell/src/App.js
```diff
@@ -28,9 +48,36 @@ const App = () => (
...
 )
 
-export default App
+App.propTypes = {
+    config: PropTypes.object,
+}
+
+const Plugin = ({ config }) => {
+    return (
+        <PluginOuterErrorBoundary>
```
tomzemp, line 62: this exists, so that an error in PluginLoader could be caught. I'm not really sure that this is ultimately necessary* as most of the logic there is just to handle the post-robot communication which is set to silently fail, so as not to impede plugin's usability if communication is not needed. If we do want an error boundary here, I guess we could reuse the same one from app-platform/adapter?

* right now, it's necessary based on how required props are being handled, but I think this should probably change

PR #789, opened 2023-03-06
fix: error in non-pwa apps [LIBS-315]

KaiVandivier:
New logic in the offline interface was causing this error: `[Offine interface] Cannot send service worker message - no service worker is controlling this page.`

This PR disables those actions in non-PWA apps and handles them in PWA apps in which the SW hasn't installed yet

### DIFF
--- a/pwa/src/offline-interface/offline-interface.js
+++ b/pwa/src/offline-interface/offline-interface.js
@@ -49,6 +49,7 @@ function testPWAAndSW() {
 
 /** Helper to simplify SW message sending */
 function swMessage(type, payload) {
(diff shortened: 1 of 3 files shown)


pwa/src/offline-interface/offline-interface.js
```diff
@@ -49,6 +49,7 @@ function testPWAAndSW() {
 
 /** Helper to simplify SW message sending */
 function swMessage(type, payload) {
+    console.log('SW Message:', { type, payload })
```
HendrikThePendric, line 52: Remove

HendrikThePendric (approved):
Left one comment

PR #793, opened 2023-03-13
fix(pwa): add precache filters [LIBS-482]

KaiVandivier:
Addresses [LIBS-482](https://dhis2.atlassian.net/browse/LIBS-482)

Doesn't address translation files yet because CRA handles those, which isn't a trivial fix unfortunately.

Tested locally by building the PWA app and verifying the files in `public/exclude-from-precache/` don't end up in the service worker's precache manifest -- you can either 1. build the app and inspect the generated `service-worker.js` file for any trace of files in the `exclude-from-precache/` dir, or you can run the `demo` script in the PWA app, visit the app, open the dev tools, inspect the workbox precache in the CacheStorage section, and verify the omitted files aren't there.

Added a few empty dummy files in the pwa app for testing.

I also tested it locally with plugins (including the [fix here](https://github.com/dhis2/app-platform/pull/792)) and verified it's working correctly.

[LIBS-482]: https://dhis2.atlassian.net/browse/LIBS-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomzemp (approved):

HendrikThePendric (approved):
LGTM, and great that you've added docs for this options too.

PR #806, opened 2023-04-27
fix: omit `moment-locales` from precache

KaiVandivier:
Intended to help with https://dhis2.atlassian.net/browse/DHIS2-15097

This saves approximately 140 files from the precache, which should speed up SW installation time significantly


pwa/src/service-worker/set-up-service-worker.js
```diff
@@ -127,10 +127,15 @@ export function setUpServiceWorker() {
...
-        const restOfManifest = precacheManifest.filter(
-            (e) => e !== indexHtmlManifestEntry
-        )
+        // Handle the rest of files in the manifest - filter out index.html,
+        // and all moment-locales, which bulk up the precache and slow down
+        // installation significantly. Handle them network-first in app shell
+        const restOfManifest = precacheManifest.filter((e) => {
+            return (
+                e !== indexHtmlManifestEntry &&
+                !e.url.startsWith('./static/js/moment-locales/')
```
Birkbjo, line 136: Would be nice to change this to a pattern-array to make it less hardcoded. Eg.
```js
MANIFEST_EXCLUDE_URL_PATTERNS = [^\.\/static\/js\/moment-locales\/] 


return (e !== indexHtmlManifestEntry &&
                e.url.startsWith('./static/js/moment-locales/') &&
               MANIFEST_EXCLUDE_URL_PATTERNS.some(p => p.test(e.url))
   )
```

Birkbjo (approved):

PR #811, opened 2023-06-27
fix: maximum call stack size exceeded at paths.js (findYarnLock function)

CelsoDeCarvalho:
> I was facing this error when I run the d2 app scripts init pmtct to create a new app. Then I found this solution in the CoP: https://community.dhis2.org/t/rangeerror-maximum-call-stack-size-exceeded-error-when-i-run-d2-app-scripts-init-my-app/49023

> But` there we needa delete findYarnLock   function. Earlier today, when a was checking this function, I noticed that this function is in an infinite loop on windows systems because the base case inside this function is not verified only in windows.

dhis2 error

`The line if(base==="/")  I think we are checking only for mac or linux OS 'cause in windows, is if(base==="C:\\") .`
Capture123

---
## I tried this solution and it worked for me.

Capturesdsd

### DIFF
--- a/cli/src/lib/paths.js
+++ b/cli/src/lib/paths.js
@@ -20,7 +20,7 @@ const findYarnLock = (base, rootDirectory) => {
 
 module.exports = (cwd = process.cwd()) => {
     const base = path.resolve(cwd)
-    const rootDirectory = path.parse(cwd).root;
+    const rootDirectory = path.parse(cwd).root
     const paths = {
         babelConfig: path.join(__dirname, '../../config/babel.config.js'),
         configDefaultsApp: path.join(
--- a/cli/src/lib/paths.js
+++ b/cli/src/lib/paths.js
@@ -1,6 +1,8 @@
 const fs = require('fs')
 const path = require('path')
 const { reporter } = require('@dhis2/cli-helpers-engine')
+const os = require('os');
+
 
 const shellSource = path.dirname(
     require.resolve('@dhis2/app-shell/package.json')
@@ -8,7 +10,10 @@ const shellSource = path.dirname(
 const shellAppDirname = 'src/D2App'
 
 const findYarnLock = (base) => {
-    if (base === '/') {
+    // Get the root path or drive based on the operating system
+    const rootPathByOS = os.platform() === 'win32' ? path.parse(process.cwd()).root : '/';
+
+    if (base === rootPathByOS) {
         return null
     }
 
--- a/cli/src/lib/paths.js
+++ b/cli/src/lib/paths.js
@@ -7,21 +7,20 @@ const shellSource = path.dirname(
 )
 const shellAppDirname = 'src/D2App'
 
-const findYarnLock = (base, directoryRoot) => {
-    if (base === directoryRoot) {
+const findYarnLock = (base, rootDirectory) => {
+    if (base === rootDirectory) {
         return null
     }
     const yarnLock = path.join(base, './yarn.lock')
     if (fs.existsSync(yarnLock)) {
         return yarnLock
     }
-    return findYarnLock(path.dirname(base), directoryRoot)
+    return findYarnLock(path.dirname(base), rootDirectory)
 }
 
 module.exports = (cwd = process.cwd()) => {
     const base = path.resolve(cwd)
-    // Get the root path or drive based on the operating system
-    const rootPathByOS = path.parse(cwd).root;
+    const rootDirectory = path.parse(cwd).root;
     const paths = {
         babelConfig: path.join(__dirname, '../../config/babel.config.js'),
         configDefaultsApp: path.join(
@@ -45,7 +44,7 @@ module.exports = (cwd = process.cwd()) => {
 
         base,
         package: path.join(base, './package.json'),
-        yarnLock: findYarnLock(base, rootPathByOS),
+        yarnLock: findYarnLock(base, rootDirectory),
         dotenv: path.join(base, './.env'),
         config: path.join(base, './d2.config.js'),
         readme: path.join(base, './README.md'),
--- a/cli/src/lib/paths.js
+++ b/cli/src/lib/paths.js
@@ -7,23 +7,21 @@ const shellSource = path.dirname(
 )
 const shellAppDirname = 'src/D2App'
 
-const findYarnLock = (base) => {
-    // Get the root path or drive based on the operating system
-    const rootPathByOS = path.parse(process.cwd()).root;
-
-    if (base === rootPathByOS) {
+const findYarnLock = (base, directoryRoot) => {
+    if (base === directoryRoot) {
         return null
     }
-
     const yarnLock = path.join(base, './yarn.lock')
     if (fs.existsSync(yarnLock)) {
         return yarnLock
     }
-    return findYarnLock(path.dirname(base))
+    return findYarnLock(path.dirname(base), directoryRoot)
 }
 
 module.exports = (cwd = process.cwd()) => {
     const base = path.resolve(cwd)
+    // Get the root path or drive based on the operating system
+    const rootPathByOS = path.parse(cwd).root;
     const paths = {
         babelConfig: path.join(__dirname, '../../config/babel.config.js'),
         configDefaultsApp: path.join(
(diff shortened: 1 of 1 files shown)


amcgee:
Thank you for this PR @CelsoDeCarvalho !  It generally looks good.  Could you change it to use `path.parse(process.cwd()).root` on all operating systems instead of only on Windows?  The same should work on *nix machines as well.

CelsoDeCarvalho:
@amcgee, Thanks for reply. I don't know if I get your question. Are you asking to treat all OS generically? I already do that  in this line of code.

'cause windows root path is the only one diferent from the other OS, so there am verifyng if the OS is windows them I pass `path.parse(process.cwd()).root` that give me windows root path, otherwise if the OS  ain't windows, the root path is the same for all that OS that is "/", then I pass "/" to the `rootPathByOs` variable.

CelsoDeCarvalho:
It is unnecessary to use `path.parse(process.cwd()).root`  for non windows  OS, because it is known that the root will be "/"

CelsoDeCarvalho:
Do you think that I needa change it?

amcgee:
Thanks Celso - thinking about this a little more, I think using `process.cwd()` actually might not always be correct.  If a directory with another root (like `D://`) is passed as the `base` directory it will never be found.  Also the way this change is currently implemented the root is computed recursively but will never change, so it would be a bit more efficient to only do this calculation once.

I would recommend instead changing [this line](https://github.com/dhis2/app-platform/pull/811/files#diff-ee530af03db059a02d7e41e72c3d67576a33a66347ce1e61706b1d357eb24e66R50) to `findYarnLock(base, path.parse(base).root)` and then updating the `findYarnLock` function to accept both `base` and `root` as parameters, and passing both when calling it recursively.

Let me know if this makes sense to you, thanks again for helping to fix this!

amcgee:
Hi @CelsoDeCarvalho thanks again for your contribution here!  Do you think you will have time to apply the changes requested?  I'd like to get this fixed, so if you don't have time I can also do it myself and merge a separate PR.  Let me know!

CelsoDeCarvalho:
Thanks @amcgee for text me. Oh! I had forgotten about this point. I'll make changes today. Thanks

It will be a pleasure to contribute to this repo.

CelsoDeCarvalho:
> Thanks Celso - thinking about this a little more, I think using `process.cwd()` actually might not always be correct. If a directory with another root (like `D://`) is passed as the `base` directory it will never be found. Also the way this change is currently implemented the root is computed recursively but will never change, so it would be a bit more efficient to only do this calculation once.
> 
> I would recommend instead changing [this line](https://github.com/dhis2/app-platform/pull/811/files#diff-ee530af03db059a02d7e41e72c3d67576a33a66347ce1e61706b1d357eb24e66R50) to `findYarnLock(base, path.parse(base).root)` and then updating the `findYarnLock` function to accept both `base` and `root` as parameters, and passing both when calling it recursively.
> 
> Let me know if this makes sense to you, thanks again for helping to fix this!

@amcgee  This make sense. The variable declaration inside a recursive function can decrease the perfomance. I have made the changes. But for the first point, can you please explain it?
Thanks.

cli/src/lib/paths.js
```diff
@@ -7,20 +7,21 @@ const shellSource = path.dirname(
...
         return yarnLock
     }
-    return findYarnLock(path.dirname(base))
+    return findYarnLock(path.dirname(base), directoryRoot)
 }
 
 module.exports = (cwd = process.cwd()) => {
     const base = path.resolve(cwd)
+    // Get the root path or drive based on the operating system
+    const rootPathByOS = path.parse(cwd).root;
```
amcgee, line 24: This is no longer dependent on the operating system, it should work for all operating systems, so we should remove this comment and change the name to simply `rootDirectory`

```suggestion
    const rootDirectory = path.parse(cwd).root;
```

amcgee:
@CelsoDeCarvalho thanks and certainly!

Basically, the variable `cwd` here will default to `process.cwd()` (the current working directory from which this command is run) but *it can be overridden with a command line argument*.  So it would be possible to do this:

```
> cd D:\test\folder
> d2 app scripts init --cwd C:\development\apps mynewapp
```

In this case, `process.cwd()` would be on the `D:\` drive but the `cwd` argument (which indicates where the new app should be created) would be on the `C:\` drive.  Since we're starting from the `cwd` directory and traversing up the file system heirarchy to find the root directory, we'd eventually get to `C:\`.  If we used `path.parse(process.cwd()).root` instead of `path.parse(base).root` it would try to match `D:\` as the eventual base condition to terminate the recursion, but that would never happen.  So we need to make sure that we're finding the root directory *of the directory we're working on* which might not be the current working directory in bash.  Does that make sense?

This change looks good to me now!  I made one very minor suggestion to update the naming of the root variable.  Thanks again for your contribution!

amcgee:
Don't worry about the commit message lint issue - it's caused by the fact that several commits here aren't exactly following conventional commit standards - [6b61c07](https://github.com/dhis2/app-platform/pull/811/commits/6b61c07557b1201f029b0b26ac6ced46f72cc984) is missing a `:` and [201b22a](https://github.com/dhis2/app-platform/pull/811/commits/201b22aaa1bab114bb40553a97284cfd58b94820) is missing a ` ` (space) after the `:`.

I will squash-merge this branch when I merge it, so no need to fix those on your branch :-)

CelsoDeCarvalho:
> @CelsoDeCarvalho thanks and certainly!
> 
> Basically, the variable `cwd` here will default to `process.cwd()` (the current working directory from which this command is run) but _it can be overridden with a command line argument_. So it would be possible to do this:
> 
> ```
> > cd D:\test\folder
> > d2 app scripts init --cwd C:\development\apps mynewapp
> ```
> 
> In this case, `process.cwd()` would be on the `D:\` drive but the `cwd` argument (which indicates where the new app should be created) would be on the `C:\` drive. Since we're starting from the `cwd` directory and traversing up the file system heirarchy to find the root directory, we'd eventually get to `C:\`. If we used `path.parse(process.cwd()).root` instead of `path.parse(base).root` it would try to match `D:\` as the eventual base condition to terminate the recursion, but that would never happen. So we need to make sure that we're finding the root directory _of the directory we're working on_ which might not be the current working directory in bash. Does that make sense?
> 
> This change looks good to me now! I made one very minor suggestion to update the naming of the root variable. Thanks again for your contribution!

Hi @amcgee 

Lemme see if I get it. Basically you saying that there will be possible to create app in a directory passed through the command line in `--cwd` flag and some times if the directory am passing be diferent of my current working directory, the recursion stop condition will not be verified rite? And we have done that point 'cause we're using cwd variable that I think this is the one that will be used as the command line argument. Please let me know if am right. Thanks

I have made the changes.

amcgee:
> > @CelsoDeCarvalho thanks and certainly!
> > Basically, the variable `cwd` here will default to `process.cwd()` (the current working directory from which this command is run) but _it can be overridden with a command line argument_. So it would be possible to do this:
> > ```
> > > cd D:\test\folder
> > > d2 app scripts init --cwd C:\development\apps mynewapp
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this case, `process.cwd()` would be on the `D:\` drive but the `cwd` argument (which indicates where the new app should be created) would be on the `C:\` drive. Since we're starting from the `cwd` directory and traversing up the file system heirarchy to find the root directory, we'd eventually get to `C:\`. If we used `path.parse(process.cwd()).root` instead of `path.parse(base).root` it would try to match `D:\` as the eventual base condition to terminate the recursion, but that would never happen. So we need to make sure that we're finding the root directory _of the directory we're working on_ which might not be the current working directory in bash. Does that make sense?
> > This change looks good to me now! I made one very minor suggestion to update the naming of the root variable. Thanks again for your contribution!
> 
> Hi @amcgee
> 
> So Lemme see if I get it. Basically you saying that there will be possible to create app in a directory passed through the command line in `--cwd` flag and some times if the directory am passing be diferent of my current working directory, the recursion stop condition will not be verified, rite? And we have done that point 'cause we're using cwd variable that I think this is the one that will be used as the command line argument. Please let me know if am right. Thanks
> 
> I have made the changes.

That's correct, thank you @CelsoDeCarvalho !

amcgee:
I manually built, tested, and linted this PR before merging.  Thanks again @CelsoDeCarvalho for the contribution!!

PR #812, opened 2023-08-01
fix: catch sw precaching errors [DHIS2-15625]

KaiVandivier:
https://dhis2.atlassian.net/browse/DHIS2-15625

Proof of concept of catching precaching errors so we can do something about it

One way is to use `patch-package` to edit a few lines in `workbox-precaching` to catch errors and prevent precaching errors from scrapping the entire installation (which happens if [`event.waitUntil()`](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil) fails in the service worker's `install` event)

Another way is to extend the `PrecacheController` object to have some more flexibility about how it's used, which may be useful depending on how we choose to respond to precaching failures

To do:
- [ ] Some testing instructions
- [ ] Remove logs and comments

### DIFF
--- a/pwa/src/service-worker/set-up-service-worker.js
+++ b/pwa/src/service-worker/set-up-service-worker.js
@@ -1,4 +1,10 @@
-import { precacheAndRoute, matchPrecache, precache } from 'workbox-precaching'
+import {
+    precacheAndRoute,
+    matchPrecache,
+    precache,
+    // PrecacheController,
+    // PrecacheRoute,
+} from 'workbox-precaching'
 import { registerRoute, setDefaultHandler } from 'workbox-routing'
 import {
     NetworkFirst,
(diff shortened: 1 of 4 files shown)


pwa/src/service-worker/set-up-service-worker.js
```diff
@@ -155,6 +161,75 @@ export function setUpServiceWorker() {
...
+            { url: './index.html', revision: '1' },
+            // The following are expected to fail
+            { url: 'https://not-a-site.com/hey.jpg', revision: '1' },
+            { url: 'https://not-a-site.com/bogus1.jpg', revision: '1' },
+            { url: 'https://not-a-site.com/derp2.jpg', revision: '1' },
+            { url: 'https://not-a-site.com/diddlibap3.jpg', revision: '1' },
+            { url: 'https://not-a-site.com/squip4.jpg', revision: '1' },
+        ])
+
+        /* An alternative to patch package, where a custom precache controller is made and implemented
```
KaiVandivier, line 177: This is included here to show an example of another approach to catching & handling precaching errors (option 2 in the main PR comment)

PR #813, opened 2023-08-10
chore: catch alpha up with master

KaiVandivier:
This should add fixes from 10.3.2 through 10.3.9, including hiding the icon of "plugin-only app" from the apps menu -- see this [slack thread](https://dhis2.slack.com/archives/C03TYMREXDL/p1691524394340579)


tomzemp (approved):

eirikhaugstulen (approved):
👍

PR #823, opened 2023-12-21
feat: support plugin wrappers [LIBS-397]

tomzemp:
This PR has changes for app-platform to allow developers to use experimental plugin wrappers (see corresponding app-runtime PR: https://github.com/dhis2/app-runtime/pull/1366)

The main change in this code is adding `<PluginLoader>` in the app shell which sets up communication with the parent window. For the communication to work, `<Plugin>` component needs to be used in the app, otherwise, the communication will fail (the plugin would still load, so you can still build a plugin with this updated app shell and not use `<Plugin>` if you do not want this communication channel, but do want a separate plugin build)

There is some additional updates in app/adapter for hoisting plugin alerts and errors.


kabaros (approved):
LGTM

PR #824, opened 2024-01-16
chore(deps): update shell to ui@9

Birkbjo:
Theoretically this should be a breaking change, since we update the major version of a dependency. However the changes in UI are not really changes that should affect people - just some removed proptypes - nevertheless they were part of the public API.

BREAKING CHANGE: Updated major version of UI


kabaros (approved):

KaiVandivier (changes requested):
I think this needs to be released as a breaking change because it will actually break builds unless the consumer changes something (upgrades the rest of the app to `ui`@9 or uses a resolution) -- here's an example of an app with a dependency on `ui@^8`:

Screenshot 2024-01-26 at 5 07 37 PM

Birkbjo:
@KaiVandivier The commit has `breaking change`, so it would be a major version bump.

KaiVandivier (approved):

PR #825, opened 2024-01-17
feat: add configurable `dir` for language directions to app adapter [DHIS2-16480]

KaiVandivier:
[DHIS2-16480](https://dhis2.atlassian.net/browse/DHIS2-16480)

This extends the CLI, App Shell, and App Adapter to set up the `dir` HTML attribute to support RTL layouts.

The "global" `dir` at the top level of the document is set to the direction that's specified by the d2.config option so that it affects modals, alerts, and other portal elements, and then the user's locale direction is passed down to the header bar.

This also includes some logic to improve the handling of locales: currently, specific locales like "Arabic (Egypt)" (`ar_EG`) which are available from the back end aren't well supported -- for one thing, we should extend our efforts in Transifex to provide translations for those. In the App Adapter, however, we can check to see if translations exist for that specific locale, and if not, try to fall back to a more general locale like `ar`. That behavior can be seen in the console of the RTL screenshots below.

It also includes fixes for Moment, which doesn't accept locales in the format used before. I also added some fallback logic there to use a similar locale when possible

**ar_EG with `direction: 'auto'`** (notice console and `dir` attributes in the DOM)
Screenshot 2024-01-17 at 5 57 06 PM

**ar_EG with default `direction` (ltr)**
Screenshot 2024-01-17 at 5 55 53 PM

**en**
Screenshot 2024-01-17 at 5 58 48 PM

[DHIS2-16480]: https://dhis2.atlassian.net/browse/DHIS2-16480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adapter/src/index.js
```diff
@@ -12,6 +12,7 @@ const AppAdapter = ({
     appVersion,
     url,
     apiVersion,
+    direction,
```
kabaros, line 15: do we need to set the default to 'ltr' here in case no value was provided?

cli/config/d2.config.app.js
```diff
@@ -1,6 +1,12 @@
 const config = {
     type: 'app',
 
+    // sets the `dir` HTML attribute for the app:
+    // options are 'ltr', 'rtl', and 'auto'. If set to 'auto', the direction
+    // will be inferred by the user's UI locale.
+    // The header bar direction will always be set by the locale.
+    direction: 'ltr',
```
kabaros, line 8: does this mean a direction has to be provided in the config, or it will fall back to ltr if not provided?

adapter/src/utils/useLocale.js
```diff
@@ -3,16 +3,45 @@ import i18n from '@dhis2/d2-i18n'
...
+    const separator = match[0] // '-' or '_'
+    const splitLocale = locale.split(separator)
+    for (let i = splitLocale.length - 1; i > 0; i--) {
+        const shorterLocale = splitLocale.slice(0, i).join(separator)
+        if (i18n.hasResourceBundle(shorterLocale, I18N_NAMESPACE)) {
+            return shorterLocale
+        }
+        console.log(`Translations for locale ${shorterLocale} not found`)
+    }
+
```
kabaros, line 39: for this logic for getting the shorter version, instead of doing it manually by splitting .. couldn't we use the Locale methods, i.e.

```
const locale = new Intl.Locale('ja-Jpan-JP');

console.log(locale.language, locale.region, locale.script);
// will give us language = ja, region = JP, script = Jpan
// and we can try to get the locale for language or `${language}-${region}``
```

adapter/src/utils/useLocale.js
```diff
@@ -23,22 +52,41 @@ const setGlobalLocale = (locale) => {
...
+    const resolvedLocale = validateLocaleByBundle(locale)
+    i18n.changeLanguage(resolvedLocale)
+
+    console.log('🗺 Global d2-i18n locale initialized:', resolvedLocale)
+}
+
+// Sets the global direction based on the app's configured direction
+// (which should be done to affect modals, alerts, and other portal elements),
+// then returns the locale's direction for use on the header bar
+const handleDirection = ({ locale, configDirection }) => {
```
kabaros, line 64: this method is doing two distinct things: getting the global direction and setting the root direction, and the name is confusing its responsibility .. maybe the root direction can be set inline in the hook, and this is just a pure function that returns the global direction.

adapter/src/utils/useLocale.js
```diff
@@ -23,22 +52,41 @@ const setGlobalLocale = (locale) => {
...
+    // for i18n.dir, need JS-formatted locale
```
kabaros, line 65: Also for all of this, maybe we can be more defensive and wrap it in a try/catch and fallback to 'ltr' - there are few external inputs here (the Jav locale) that might go wrong so it'd be good to make sure it doesn't crash.

adapter/src/utils/useLocale.js
```diff
@@ -23,22 +52,41 @@ const setGlobalLocale = (locale) => {
...
+    const jsLocale = transformJavaLocale(locale)
+    const localeDirection = i18n.dir(jsLocale)
```
kabaros, line 67: Finally could we perhaps split all this part (getGlobalDirection + transformJavaLocale + i18n.dir) into another file .. just keep useLocale simple

adapter/src/utils/useLocale.js
```diff
@@ -23,22 +52,41 @@ const setGlobalLocale = (locale) => {
     }
     moment.locale(locale)
 
-    const simplifiedLocale = simplifyLocale(locale)
-    i18n.changeLanguage(simplifiedLocale)
+    const resolvedLocale = validateLocaleByBundle(locale)
```
kabaros, line 55: maybe a better name for `validateLocaleByBundle` is just `getResolvedLocale` .. that's what it does for me (the validation is an internal detail)

adapter/src/utils/useLocale.js
```diff
@@ -47,16 +95,21 @@ const settingsQuery = {
...
+        configDirection,
+    })
 
     if (error) {
         // This shouldn't happen, trigger the fatal error boundary
         throw new Error('Failed to fetch user locale: ' + error)
     }
 
-    return { loading: loading || !locale, locale }
+    return { loading: loading || !locale, locale, direction }
```
kabaros, line 114: it would be nice to add a test for this hook .. I know it's not particularly part of this PR since it didn't exist before, but there is quite a bit of logic here and having a test to validate all the permuations would be nice

cli/config/d2.config.app.js
KaiVandivier: These are the config defaults, so it will use 'ltr' if nothing is provided 👍

adapter/src/index.js
KaiVandivier: I suppose that would be good, in case something goes wrong with the env var -- which makes me wonder, should we handle 'default to LTR' here only, and leave out the default env var? 🤔 It would tidy up the env vars slightly in the default case

adapter/src/utils/useLocale.js
KaiVandivier: Hmm yeah that's a nice helper function -- if that works, I think that can help make that code look a lot more semantic 👍 Implementing it is a bit complex given the existing translation files we have to work with, for example `uz_UZ_Cyrl` isn't in the order that `Intl.Locale` wants, but I'll see what I can do (when using hyphens, `new Intl.Locale('uz-UZ-Cyrl')` throws an error)

Here's a screenshot from the Dashboard app's locales:

Screenshot 2024-01-19 at 1 42 58 PM

adapter/src/utils/useLocale.js
KaiVandivier: That may take me some time since I'm a bit out of practice with mocking things like `useDataQuery` and `i18n.dir`, but I'll see what I can do 👍

adapter/src/utils/useLocale.js
KaiVandivier: It's possible `i18n.dir` is already quite robust in that way -- you can pass it any string (or an array even, lol), and it will default to `'ltr'`

Was there something else here you think is fragile?

adapter/src/index.js
kabaros: yeah I would say so .. it would keep the default env vars tidier

adapter/src/utils/useLocale.js
kabaros: I guess I had `locale` being undefined in mind .. I thought it might be on first-render or something, and it's better to be safe. But maybe a try/catch around the whole useLocale effect is even safer (the stuff with splitting locale strings could throw as well in a couple of places)

adapter/src/utils/useLocale.js
kabaros: ok, we can delay it for later too .. but I didn't mean to test the whole thing, just the useLocale hook with something like [react-hooks-testing-library](https://github.com/testing-library/react-hooks-testing-library)

KaiVandivier: Ah okay, just the `useLocale` hook will be quite a bit simpler -- this comment was placed in a different hook though 😁

cli/config/d2.config.app.js
KaiVandivier: (Removed to keep default env vars tidier)

adapter/src/utils/useLocale.js
KaiVandivier: In my recent changes, I approached error handling in a bit more specific way at each step with useful fall-backs -- my idea is that if one step doesn't work, the other steps can still succeed and provide useful localization. Not many things in here can actually throw, but those things are wrapped in try/catch

adapter/src/utils/useLocale.js
KaiVandivier: Tests added

adapter/src/utils/localeUtils.js
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * @param {UserSettings} userSettings
+ * @returns Intl.Locale
+ */
+export const parseLocale = (userSettings) => {
+    try {
+        // proposed property
+        if (userSettings.keyUiLanguageTag) {
+            return new Intl.Locale(userSettings.keyUiLanguageTag)
+        }
```
KaiVandivier, line 38: This is the new property Austin and I proposed, as a proof of concept of forward and backward compatibilty

adapter/src/utils/localeUtils.js
```diff
@@ -0,0 +1,162 @@
...
+    console.log('🗺 Global d2-i18n locale initialized:', finalLocaleString)
+}
+
+/**
+ * Moment locales use a hyphenated, lowercase format.
+ * Since not all locales are included in Moment, this
+ * function tries permutations of the locale to find one that's supported.
+ * NB: None of them use both a region AND a script.
+ * @param {Intl.Locale} locale
+ */
```
KaiVandivier, line 116: New fixes in here to Moment locale loading since the last review, and also split to its own function

adapter/src/utils/localeUtils.js
```diff
@@ -0,0 +1,162 @@
...
+    // worst-case fallback
+    return new Intl.Locale(window.navigator.language)
+}
+
+/**
+ * Test locales for available translation files -- if they're not found,
+ * try less-specific versions.
+ * Both "Java Locale.toString()" and BCP 47 language tag formats are tested
+ * @param {Intl.Locale} locale
+ */
```
KaiVandivier, line 58: Reworked the below take advantage of `Intl.Locale` semantics

kabaros (approved):

PR #828, opened 2024-02-14
fix: remove >0 check so last array item can be removed too

HendrikThePendric:


adapter/src/components/__tests__/Alerts.test.js
```diff
@@ -49,9 +49,10 @@ describe('Alerts', () => {
 
     it('can hide an alert', async () => {
         const msg = 'TEST ALERT'
+        const options = { permanent: true }
```
HendrikThePendric, line 52: This ensures the alert does't auto-hide. We need this to verify that hiding programmatically does indeed clean up the DOM.

Birkbjo (approved):
LGTM.

PR #835, opened 2024-03-15
feat: add redirect to safe login mode for login apps

tomzemp:
Adds login-app-specific error boundary that shows button to redirect to "safe" login page:

image

### DIFF
--- a/adapter/src/components/ErrorBoundary.js
+++ b/adapter/src/components/ErrorBoundary.js
@@ -63,11 +63,13 @@ export class ErrorBoundary extends Component {
 
     handleSafeLoginRedirect = () => {
         window.location.href =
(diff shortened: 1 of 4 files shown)


adapter/src/components/ErrorBoundary.js
```diff
@@ -61,8 +61,14 @@ export class ErrorBoundary extends Component {
         })
     }
 
+    handleSafeLoginRedirect = () => {
+        window.location.href =
+            this.props.url + '/dhis-web-commons/security/login.action'
```
kabaros, line 66: I think this prop would be btter named `baseUrl`? and maybe be more defensive with the url construction here, in case the base URL has a trailing slash, to avoid a double slash

kabaros (approved):

PR #837, opened 2024-03-21
feat: enable publishing login_app apps

kabaros:


tomzemp (approved):
👍
There's also already a helper function for this, so requiring that (`const { isApp } = require('../lib/parseConfig')`) and then using it `if (!isApp(appConfig.type)) {` would also work

PR #843, opened 2024-04-10
fix(deps): use exact version of workbox to avoid broken 6.6.1 version

KaiVandivier:
**Update:** tested this out on alpha, and it's not successful because `react-scripts` still has a dependency on `workbox-webpack-plugin@^6.4.1`, which resolves to the broken 6.6.1 upon installation of the CLI

---

Currently, new installations of `cli-app-scripts` end up throwing an error on Node 14:
```
error workbox-webpack-plugin@6.6.1: The engine "node" is incompatible with this module. Expected version ">=16.0.0". Got "14.21.3"
```
This seems to be due to a broken release of `workbox` that should have been a major version, but it gets installed because of the ^6 dependency in the platform

The 6.6.1 release is now flagged as `deprecated` when installed on Node 16, and seems to be removed from the [release history](https://github.com/GoogleChrome/workbox/releases), which also shows the next major release as having the exact same breaking change in question (v7: Node 16 is now the minimum version). There is also a reported [bug](https://github.com/GoogleChrome/workbox/issues/3220)

This will avoid increasing the minimum Node requirement to 16 for app-platform

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -67,8 +67,8 @@
         "terser-webpack-plugin": "^5.3.1",
         "webpack": "^5.41.1",
         "webpack-dev-server": "^4.7.4",
-        "workbox-build": "^6.1.5",
-        "workbox-webpack-plugin": "^6.5.4"
+        "workbox-build": "6.5.4",
+        "workbox-webpack-plugin": "6.5.4"
     },
     "bin": {
         "d2-app-scripts": "./bin/d2-app-scripts"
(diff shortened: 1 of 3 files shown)


KaiVandivier:
Tested on alpha, and this isn't a successful solution because `react-scripts` still has a dependency on `workbox-webpack-plugin@^6.4.1`, which resolves to the broken 6.6.1 upon any new installation of `cli-app-scripts` unfortunately

PR #845, opened 2024-04-23
feat: enable login app type

tomzemp:
This PR merges in the changes that are on alpha v10.6.0-alpha3 release (https://github.com/dhis2/app-platform/releases/tag/v10.6.0-alpha.3) that make possible using app-platform for login app types.


kabaros (approved):

tomzemp:
Pushed some clean up (https://github.com/dhis2/app-platform/pull/845/commits/837624f4cb2d64438fcd612c7bdc7f8cbff0da34)

Based on comments from @KaiVandivier :
> 1. It might be nice to choose one of either login or loginApp as the boolean prop used through the adapter
>2. I didn’t really understand this comment in the ServerVersionProvider: “version is not currently available, minimum for login app” — is that like “we don’t need the whole system info for the login app, but we need the server version at the minimum to proceed”?
P.S. since the login app type uses skipApiVersion on the app-runtime provider, is fetching the server version in a login app just to satisfy the type requirement for the config prop on the Provider or something (or the version parsing functions)?

1. harmonized
2. these are dated given that the changes were initially made before versioning was available and supported on login configuration

PR #846, opened 2024-04-25
fix: handle new login endpoints [LIBS-600]

tomzemp:
See https://dhis2.atlassian.net/browse/LIBS-600

## Summary

This change fixes the app-platform login modal to:
1. use new auth/login endpoints if v41 or greater
2. use old struts login (/dhis-web-commons-security/login.action) if v40 or less

## Detecting version

If a server is v41 or greater, api/loginConfig is accessible with the version. Therefore, this PR assumes that if a request to api/loginConfig succeeds that the server is v41+. This logic, obviously, has some limitations because the request could fail for other reasons (e.g. server is down), but this is not in general likely.

### General
I've disabled the log in button when login is in progress.

### Normal apps
When using api/auth/logic, we have a REST api and will get a 400-range error back if authentication fails. This means that we can catch errors and display them back to the user. When using the struts endpoint, even successful authentication errors out. As a consequence, I've added error display if the server is detected as v41+ and have otherwise used the previous approach of refreshing the window (for v40 or less).

Note that you can successfully authenticate (regardless of version), and have CORS errors that are only detected on reload (note that this appears like the older version workflow: you authenticate, the app reloads, and then the cors error leads to the login modal to display again). I have not made any modifications to specifically catch CORS errors and display them.

v42:
login_modal_dev

v39:
login_modal_239dev

### Login app 
I have added some handling for the case where you are have a login app. Previously, running the login app required that you specify the `DHIS2_BASE_URL` environment variable. This adds a bit more refactoring to the code, but I think it's worth it as it makes login app development less finicky and also would be relevant if we introduce "public" app types in the future.

image

## Code notes

I thought about doing a bit more a refactor (e.g. to get rid of need for page refreshing with v41+, but then I felt that was a bit complicated and not worth the effort at the moment).


adapter/src/index.js
```diff
@@ -23,6 +23,7 @@ const AppAdapter = ({
     loginApp,
     children,
 }) => {
+    const [loginBaseUrl, setLoginBaseUrl] = useState(url)
```
tomzemp, line 26: We need to pass the base url (logic in ServerVersionProvider) to LoginAppWrapper in order to display the correct redirection for safe mode login. 

The easiest option would be to just move the `LoginAppWrapper` component into `ServerVersionProvider`, but I thought maybe the current code structure was preferable in terms of displaying where the wrapper was? So hence this use of state variable to track the updates from `ServerVersionProvider` in this parent component.

adapter/src/components/LoginModal.js
```diff
@@ -6,57 +6,133 @@ import {
...
+                    {!loginApp
+                        ? i18n.t('Please sign in')
+                        : i18n.t('Specify server')}
+                </ModalTitle>
 
                 <ModalContent>
+                    {error && (
+                        <div className="errorNotification">
+                            <NoticeBox error title={i18n.t('Could not log in')}>
+                                {error?.message || error?.details?.message}
```
tomzemp, line 132: we actually seem to not be throwing the error so that it can be properly caught here...so I could update that or we could just skip displaying a message for now

adapter/src/index.js
Birkbjo: Wait, why can't we just read from the server-version context in `LoginAppWrapper`, since it's a child of that provider?

tomzemp: @Birkbjo : we can use `useConfig` within `LoginAppWrapper` to get the baseUrl. I didn't want to do that initially because I thought it was kind of confusing to use the hook from app-runtime there, but we're also using `useConfig` in the `ConnectedHeaderBar` component, so I guess using `useConfig` in this case would be more consistent. I think my hesitation was just that within the directory we have some components that have access to the context and some that do not, so you have to pay attention to where these components are in the hierarchy.

I've updated to use `useConfig` in `LoginAppWrapper`. 🙏

adapter/src/components/LoginModal.js
KaiVandivier: I think it would be nice to show a message! I played around a little to see what would be needed, and I think it just needs this at line 17 in `api.js`:
```diff
-                    reject('Request failed ' + response.statusText)
+                    response.json().then(json => {
+                        reject(json)
+                    })
```

adapter/src/components/LoginModal.js
```diff
@@ -6,57 +6,133 @@ import {
...
-                )
-            } catch (e) {
-                console.log(
-                    'TODO: This will always error and cancel the request until we get a real login endpoint!'
-                )
-            }
 
-            // TODO: Hacky solution... this shouldn't require a reload
-            window.location.reload()
+            const useNewLoginAPI = await getUseNewLoginAPI(server)
```
KaiVandivier, line 102: This feels like it's clashing a bit with the react hook naming conventions -- what do you think about `isNewLoginAPIAvailable` and `getIsNewLoginAPIAvailable`?

KaiVandivier (approved):
Nice! This looks great, I like the changes you added 😀 

Added a few tips for your consideration 🙂

adapter/src/components/LoginModal.js
KaiVandivier: lol the message from the backend feels slightly harsh

Screenshot 2024-04-26 at 11 36 00 AM

adapter/src/components/ServerVersionProvider.js
```diff
@@ -192,6 +194,7 @@ export const ServerVersionProvider = ({
             plugin={plugin}
             parentAlertsAdd={parentAlertsAdd}
             showAlertsInPlugin={showAlertsInPlugin}
+            skipApiVersion={loginApp ? true : false}
```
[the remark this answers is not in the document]
tomzemp: 🤦‍♂️

adapter/src/components/LoginModal.js
```diff
@@ -6,57 +6,133 @@ import {
...
-export const LoginModal = ({ appName, baseUrl }) => {
+const getUseNewLoginAPI = async (server) => {
+    try {
+        // if loginConfig is available, the instance can use new endpoints
+        await get(`${server}/api/loginConfig`)
+        return true
+    } catch (e) {
+        // if loginConfig is not available, the instance must use old endpoints
+        return false
+    }
```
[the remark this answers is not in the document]
tomzemp: I'm going to `console.error` the error in case it's something beside it just doesn't exist (or if we get an error back on an instance that should have it)

adapter/src/components/LoginModal.js
tomzemp: @KaiVandivier:  how about 
```
                if (response.status !== 200) {
                    response
                        .json()
                        .then((json) => {
                            reject(json)
                        })
                        .catch(() => {
                            reject('Request failed ' + response.statusText)
                        })
                    return
                }
```

I think there could be api requests returning non-json responses, but the json() method should throw an error and then we can catch and reject with `response.statusText` like before?

KaiVandivier: Good call! That sounds perfect 👌

PR #848, opened 2024-05-27
feat: start plugin and app separately [LIBS-391] [LIBS-392]

KaiVandivier:
Addresses [LIBS-391](https://dhis2.atlassian.net/browse/LIBS-391) and [LIBS-392](https://dhis2.atlassian.net/browse/LIBS-392):
1. Runs a dev build with `yarn start` instead of a minified production build -- this was fixed by allowing multiple chunks to be made in the dev build
2. Allows starting the app and the plugin separately -- the default behavior is still the same to avoid a breaking change, but you can now run `yarn start --app` to start just the app or `yarn start --plugin` to start just the plugin
    1. With the Vite changes, I think I will change the default to be to start just the app, then a user can pass the plugin flag to start just the plugin. It doesn't seem necessary to run both

Minified before & after:
minified-before
minified-after

New start script options:
new-start-script-options

[LIBS-391]: https://dhis2.atlassian.net/browse/LIBS-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIBS-392]: https://dhis2.atlassian.net/browse/LIBS-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/config/plugin.webpack.config.js
+++ b/cli/config/plugin.webpack.config.js
@@ -103,7 +103,7 @@ module.exports = ({ env: webpackEnv, config, paths }) => {
             path: paths.shellBuildOutput,
             filename: isProduction
                 ? 'static/js/plugin-[name].[contenthash:8].js'
-                : 'static/js/plugin.bundle.js',
+                : 'static/js/plugin-[name].bundle.js',
             chunkFilename: isProduction
                 ? 'static/js/plugin-[name].[contenthash:8].chunk.js'
                 : 'static/js/plugin-[name].chunk.js',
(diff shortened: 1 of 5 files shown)


tomzemp (approved):
looks good to me! 🚀

should we also update the documentation for the start command: https://github.com/dhis2/app-platform/blob/master/docs/scripts/start.md ?

PR #849, opened 2024-05-28
feat: parse pluginType from d2 config to add to manifest.webapp

KaiVandivier:
Reads a `pluginType` string from `d2.config.js` and adds it to `manifest.webapp` as `plugin_type` (see the [backend code](https://github.com/dhis2/dhis2-core/blob/096d61279a97f31219eb4b03a88aba2e8709041f/dhis-2/dhis-api/src/main/java/org/hisp/dhis/appmanager/App.java#L225-L229) that parses `manifest.webapp`)

`pluginType: 'DASHBOARD'` shows up in dashboard app list:
plugin-type-dashboard

`pluginType: 'CAPTURE'` doesn't show up in dashboard app list:
plugin-type-capture

config:
pwa-plugin-type-config


tomzemp (approved):
looks good!

Do you think it might be appropriate to do convert pluginType to upper case, so that we don't end up people with uploading variants like `dashboard` or `Dashboard`, which will then be unique types? I don't think there's a good reason to allow for case-sensitive definition of plugin types, and doing the conversion on manifest generation would probably prevent some unintended errors.

KaiVandivier:
Good call... I think I'll implement that as a validation step, so that the user gets feedback that it should be in all caps, so they know what to expect if they try to filter for it

PR #852, opened 2024-06-14
fix: do not encode username, password [LIBS-625]

tomzemp:
This fixes an issue where the username and password were being encoded before being passed to api/login. 

I had copied some of the code from the old login logic, but that is using `Content-Type:application/x-www-form-urlencoded`, so requires encoding, whereas the post with json body, does not need that encoding 🤦‍♂️.

**Before**
payload like `{"username": "admin", "password":  "District1%23"}`

**After**
payload like `{"username": "admin", "password":  "District1#"}`


KaiVandivier (approved):
Tested it out; looks good! 🙂

PR #855, opened 2024-06-20
chore: update to latest dhis2 ui

flaminic:
update dhis2 ui minimum version

### DIFF
--- a/adapter/package.json
+++ b/adapter/package.json
@@ -40,7 +40,7 @@
     "peerDependencies": {
         "@dhis2/app-runtime": "^3.10.4",
         "@dhis2/d2-i18n": "^1",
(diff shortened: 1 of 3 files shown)


tomzemp (approved):
FYI for context: 
- @kabaros asked that we update the UI dependency here because of a breaking change in temporal (https://dhis2.atlassian.net/browse/LIBS-543). 

The update here also picks up:
- the changes Flaminia made for the SharingDialog types (https://github.com/dhis2/ui/pull/1515)
- and fixes some regressions that came up with accessibility changes.

tomzemp:
PS @flaminic : I think we should also update the peer dependency in app-wrapper: https://github.com/dhis2/app-platform/blob/master/adapter/package.json#L43

PR #856, opened 2024-06-21
feat: cleanup plugin error boundary [UX-136]

tomzemp:
This PR implements styling for plugin error boundaries (see notes on https://dhis2.atlassian.net/issues/UX-136)

**Style note on screenshots**: note that the teal background / dotted blue line is styling from the app (that is the styling _around_ the plugin). The actual plugin error boundaries are the the grey boxes 9

General note: Off-ticket, I think Joe and I discussed that it might be better to not vertically center the plugin error message (because there's a greater likelihood that you would not see the error in the plugin if it's centered within the iframe). If we did actually want vertically centering, that's a small edit from this.

### Default plugin error:

This is the error that you would get if something goes wrong within the plugin's code.

**Before**
image

**After**

image

-Note: The parent iframe needs to enable write access to the clipboard in order to copy the error to the clipboard. I've submitted a [PR](https://github.com/dhis2/app-runtime/pull/1382) in app-runtime to make the iframe rendered by <Plugin> component include that access. If you implement your own wrapper, you'd need to provide that access yourself.

-To do: the ticket discusses passing the name of the plugin within the error. The premise for that was that the plugin name would be the shortName provided to the app-runtime <Plugin> component for looking up iframe src. I think before adding this shortName, we should revist the actual implementation of the plugin lookup, as I think most people are just passing urls for the src at the moment.

### Outer plugin error:

This is the error that you would get if something went wrong with the app-shell logic for loading the plugin. It should (hopefully 😅) be very unlikely that you would end up in this error state, so I've just left this error as a simple "something went wrong" rather than including/refactoring all the additional code for copying error info / trying again.

**Before:**
image

**After:**
image

### DIFF
--- a/adapter/i18n/en.pot
+++ b/adapter/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2024-04-24T13:58:13.591Z\n"
-"PO-Revision-Date: 2024-04-24T13:58:13.591Z\n"
+"POT-Creation-Date: 2024-06-21T08:27:55.991Z\n"
+"PO-Revision-Date: 2024-06-21T08:27:55.991Z\n"
 
 msgid "Save your data"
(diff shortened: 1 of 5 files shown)


KaiVandivier (approved):
Looks good 🙂

PR #857, opened 2024-06-24
fix: small text change in changelog

flaminic:


KaiVandivier (approved):
[This is intended to trigger a release for the `chore` commit to bump the UI version that was recently merged]

PR #861, opened 2024-07-06
fix(deps): update i18next-scanner version to support old plurals format again [LIBS-641]

KaiVandivier:
See [LIBS-641](https://dhis2.atlassian.net/browse/LIBS-641) and its related issue for more context (tl;dr: `i18next-scanner` has a dependency on `i18next` with `*` as the version, and `i18next` v21 broke it. Version 3.0.1+ of `i18next-scanner` adapts to the changes and makes it backwards-compatible, fixing the break it introduced for us)

We’re currently on v2, and the latest version is v4

* looking at the documentation, the breaking change in v3 of i18next-scanner  is just increasing the minimum node version to 10
* as far as I can tell, there isn't really a breaking change in v4, since an [error was introduced and then rolled back](https://github.com/i18next/i18next-scanner/issues/177#issuecomment-1178818448) – either way, it works in testing
* `i18next-scanner` now also seems to [support the new plural format too](https://github.com/i18next/i18next-scanner/releases/tag/v4.4.0) (called "compatibilityJSON v4"), when we’re ready to migrate to that

Here are some test plurals to make sure they're extracted correctly: (I added them to the simple example app, then ran `yarn start` to extract the strings, generate the translation JSON files, and start the app -- the plurals also worked when viewing the app in the browser)
Screenshot 2024-07-06 at 8 23 05 PM

[LIBS-641]: https://dhis2.atlassian.net/browse/LIBS-641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -52,7 +52,7 @@
         "html-webpack-plugin": "^5.5.0",
         "http-proxy": "^1.18.1",
         "i18next-conv": "^9",
-        "i18next-scanner": "^2.10.3",
+        "i18next-scanner": "^4.5.0",
         "inquirer": "^7.3.3",
         "lodash": "^4.17.11",
         "mini-css-extract-plugin": "^2.5.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2175,6 +2175,13 @@
     minimatch "^3.1.2"
     strip-json-comments "^3.1.1"
 
+"@gulpjs/to-absolute-glob@^4.0.0":
+  version "4.0.0"
+  resolved "https://registry.yarnpkg.com/@gulpjs/to-absolute-glob/-/to-absolute-glob-4.0.0.tgz#1fc2460d3953e1d9b9f2dfdb4bcc99da4710c021"
+  integrity sha512-kjotm7XJrJ6v+7knhPaRgaT6q8F8K2jiafwYdNHLzmV0uGLuZY43FK6smNSHUPrhq5kX2slCUy+RGG/xGqmIKA==
+  dependencies:
+    is-negated-glob "^1.0.0"
+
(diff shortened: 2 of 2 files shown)


tomzemp (approved):
I tested and agree that this works to extract strings with `{{count}}` in them and also picks up the `defaultValue` and `defaultValue_plural` 🎉

I find it a bit weird that i18next's plural updates broke the i18next scanner, given that their [documentation](https://www.i18next.com/translation-function/plurals) for the current plural handling still use the `{{count}}` in their strings 🙃. But glad they fixed it!

I also can't get the new format, e.g. `defaultValue_other` to populate correctly into our .pot files, though I would guess that has to do with our old i18next-conv dependency (we're on [major version 9](https://github.com/dhis2/app-platform/blob/2c7b7964af244116270906f21fcf82edec827a72/cli/package.json#L54), and the package seems to now be on major version 15). So maybe they've now updated to use the Intl.PluralRules terminology too?

Looking at that, I did also see that we seem to be using some default config with i18next-conv that seems to result in plural interpretation that doesn't seem to let us provide more differentiated plurals other than 1 vs plural (`"Plural-Forms: nplurals=2; plural=(n != 1)\n"` in the generated po file). So I assume we also want to update that to allow us to define more nuanced plurals (e.g. for Arabic, Russian) where that's relevant.

KaiVandivier:
Thanks for looking into that @tomzemp! I added your comments to a ticket for the follow-up work to support the new plurals format: https://dhis2.atlassian.net/browse/LIBS-643

PR #865, opened 2024-07-22
chore: set dir to auto for new apps [LIBS-645]

tomzemp:
See https://dhis2.atlassian.net/jira/software/c/projects/LIBS/issues/LIBS-645

This PR updates the configuration for newly initialized app to set `dir: 'auto'` so that by default apps will follow direction of user language, and hence display 'rtl' for Arabic, Urdu, etc.,

(After this is merged, https://github.com/dhis2/cli dependencies need to be updated)

### DIFF
--- a/cli/config/d2.config.app.js
+++ b/cli/config/d2.config.app.js
@@ -4,6 +4,7 @@ const config = {
     entryPoints: {
         app: './src/App.js',
     },
+    dir: 'auto',
 }
 
 module.exports = config


KaiVandivier (changes requested):
This would have breaking side effects for existing apps unfortunately, since this file is doing double-duty as the boilerplate file that's copied into new apps _and_ as defaults that are [read](https://github.com/dhis2/app-platform/blob/47c9dd134b42a807d78459bf34ea45bccab3bbb4/cli/src/lib/parseConfig.js#L115) from [here](https://github.com/dhis2/app-platform/blob/47c9dd134b42a807d78459bf34ea45bccab3bbb4/cli/src/lib/paths.js#L26-L29) then [applied to existing apps' config](https://github.com/dhis2/app-platform/blob/47c9dd134b42a807d78459bf34ea45bccab3bbb4/cli/src/lib/parseConfig.js#L74-L75)...

In the Vite for plugins branch though, I have some work to decouple the boilerplate and the config defaults (since applying an app entrypoint to an 'app' with only a plugin entrypoint makes things complicated) -- maybe we could point this one at `alpha` once that one is merged?

tomzemp:
> This would have breaking side effects for existing apps unfortunately, since this file is doing double-duty as the boilerplate file that's copied into new apps _and_ as defaults that are [read](https://github.com/dhis2/app-platform/blob/47c9dd134b42a807d78459bf34ea45bccab3bbb4/cli/src/lib/parseConfig.js#L115) from [here](https://github.com/dhis2/app-platform/blob/47c9dd134b42a807d78459bf34ea45bccab3bbb4/cli/src/lib/paths.js#L26-L29) then [applied to existing apps' config](https://github.com/dhis2/app-platform/blob/47c9dd134b42a807d78459bf34ea45bccab3bbb4/cli/src/lib/parseConfig.js#L74-L75)...

🤦‍♂️ Thanks.

Sounds good to address later if the vite changes will make this easier. I'm going to close for now 🙏

PR #866, opened 2024-07-22
feat: update initial app code [LIBS-644]

tomzemp:
See https://dhis2.atlassian.net/jira/software/c/projects/LIBS/issues/LIBS-644

I've updated to
- use `useDataQuery` rather than `<DataQuery>`
- added `i18n.t` wrapping on the error and loading text (I changed `...` to `Loading...`, so it doesn't look like we're just translating ellipses. I suppose that might be a valid translation as I don't know if other languages use ellipses in the same way, but it seems a bit confusing). I have changed this just to display best practice of translating strings.
- I've done a conditional return with error/loading that does not include the `<div className={classes.container}>` element. I've just done this for readability / mimic what we generally do in our apps. I thought repeating the div would be distracting as would refactoring to make it a common wrapper.
- added some optional chaining to `data?.me?.name`. Again, I've done this mostly as I think it's useful practice to model this as it helps prevent unnecessary crashes when something gets returned incorrectly.

(After this is merged, https://github.com/dhis2/cli dependencies need to be updated)

### DIFF
--- a/cli/config/init.entrypoint.js
+++ b/cli/config/init.entrypoint.js
@@ -1,4 +1,4 @@
-import { DataQuery } from '@dhis2/app-runtime'
+import { useDataQuery } from '@dhis2/app-runtime'
 import i18n from '@dhis2/d2-i18n'
 import React from 'react'
 import classes from './App.module.css'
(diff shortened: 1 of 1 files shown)


KaiVandivier:
Nice! Thanks for doing this 🙌 

I have reservations about the optional chaining -- I'd argue this isn't the best practice, because it can end up causing quiet failures that can be overlooked and become hard to diagnose, especially if a value is passed to other functions or components, where it becomes harder to identify where the problem stems from

I think a better practice is to handle the loading and error states, because problems after that would probably be code errors, where it's helpful to throw descriptive errors and "fail early", where it's easier to catch in development

In this particular case,
1. If `!error && !loading`, then `data` should be defined
2. If `data` is defined, so should `data.me` because of how the query object is constructed
So I'd advocate for using `data.me.name` here

KaiVandivier (changes requested):
Requesting changes because this should be a `feat` since it changes the behavior of `d2-app-scripts init`

tomzemp:
> I have reservations about the optional chaining -- I'd argue this isn't the best practice, because it can end up causing quiet failures that can be overlooked and become hard to diagnose, especially if a value is passed to other functions or components, where it becomes harder to identify where the problem stems from

I clearly like quiet failures more 😄, but it's a bit of a nuanced conversation and depends on whether the app is likely to be fixed or not if it crashes.

In this case, I agree, that the code should not need the optional chaining ... unless something goes wrong / changes with useDataQuery. If it were a real app, and the only purpose was to display a name to the user, I'd argue this was a fine case to have the quiet failure, so user could otherwise use the app. Obviously, though, if useDataQuery starts behaving incorrectly, it would be nice if we noticed that in the code we initialize our apps with 😜

KaiVandivier (approved):

KaiVandivier:
Yeah I see you're point -- maybe it's best evaluated on a case-by-case basis

And yeah, catching a breaking change in `useDataQuery` as soon as possible would be helpful 😅

KaiVandivier:
Merged this so I can pull it into alpha 😊 (I need to make it create a `.jsx` file there)

PR #868, opened 2024-07-31
feat: create initial TS template

kabaros:
implements [LIBS-594](https://dhis2.atlassian.net/browse/LIBS-594)

#### Implementation Details:
A templates directory containing the init files directories for both the JS and TS templates has been added. Files like `package.json`, `App.module.css`, and `package.json` don't change for either of them hence they are contained in the common directory. 

The TS template contains all the same init files as the JS template with a few more files for TS configuration, i.e:
1. **tsconfig.json**
2. **declaration files**: maintaining 2 separate declaration files under the types directory, i.e. the global.d.ts and modules.d.ts. (Merging them results in the css modules declaration being invisible).
3. **eslint.config.js** 
- This is an initial configuration file that a user can build upon after setting up the project. 
- To test this, a user can add a custom lint command to the project scripts in the package.json, for example: `"lint": "eslint ."` and run `yarn lint`. Or just run 'npx eslint .' directly in the current project directory. 
- Alternatively, there is also the option of using the `dhis2-cli-style` configurations. Start by installing it with `yarn add @dhis2/cli-style --dev` and then setting up the scripts in your package.json as shown [here](https://github.com/dhis2/cli-style/blob/master/docs/getting-started.md#setup-scripts). 
You could also adjust the eslint.config.js file to add the configurations provided by d2 styles. (To be tested: d2 imported config may not be compatible with the flat config system) 

         import config from '@dhis2-cli-style'
         import js from '@eslint/js'
         import tseslint from 'typescript-eslint'
    
         export default [
            config.eslintReact,
            js.configs.recommended,
            ...tseslint.configs.recommended,
            { // other configurations go here }
        ]

#### How to test:
- Run the init script to create a new TS project. For example, within the app-platform directory, you could run 
                           
      cli/bin/d2-app-scripts init test-ts-project --typescript
                                                  
- Change into your newly created project to test the application. Because of changes made to the start script as well, you could run the start script directly from the app-platform project. For example, following the `test-ts-project` created above:

      cd test-ts-project 
      ../cli/bin/d2-app-scripts start

[LIBS-594]: https://dhis2.atlassian.net/browse/LIBS-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kabaros:
@d-rita  to test this, I run `./app-platform/cli/bin/d2-app-scripts init test-project --typescript` (replace `./app-platform` with path to where you have the app-platform project). This will create a new `test-project` in the current directory.

For `yarn start` to work in the newly bootstrapped project, there are some changes to `start` command that need to be incorporated. Since these changes are not published yet, I use [yalc](https://github.com/wclr/yalc) to publish from `/app-platform/cli` 

image

then install this version into the bootstrapped project with `yalc add @dhis2/cli-app-scripts@12.0.0-alpha.5`, then `yarn start` should work

image

maybe @KaiVandivier has a better workflow for testing changes to cli and init/start scripts?

KaiVandivier:
Good progress here! I'm excited to see this in action 🙂 

For the testing workflow, I haven't actually had to test out any work that has involved changes to both `init` _and_ a `start` or `build` command, so don't actually have a workflow planned for this case -- your looks good though! I can share some other options for your consideration too:
1. You can init a new app from outside the repo like you described above: `./app-platform/cli/bin/d2-app-scripts init test-project --typescript`
    1. (It probably helps to install the latest alpha in the new app here too)
    2. Then you can run the CLI scripts locally using relative paths in the same way: `../app-platform/cli/bin/d2-app-scripts start` (can update the scripts in `package.json`) -- this may save running `yalc publish/push`, though I'm not too familiar with `yalc`
2. You can also init an app the same way, then run a script to copy `app-platform/cli` into `test-project`'s `node_modules/@dhis2/cli-app-scripts` (and delete `.d2/shell/node_modules/.cache`, if needed) -- then regular `yarn d2-app-scripts start` should work
3. I often end up using the example apps to test CLI changes -- to test a newly initialized app, it would also be possible to run `yarn d2-app-scripts init test-project` in the monorepo root (which will use the local `d2-app-scripts`), then add the newly initialized app to the repo workspaces. When the `@dhis2/cli-app-scripts` version in the new app matches the local package version in the workspace, it should use the local one

Here are some other, mostly subjective tips:
1. For the init files, my instinct would be to keep them all in one `init` dir instead of having a dir for JS and one for TS, since they’re already copied one-by-one instead of as a batch (e.g. a new app only gets one of the two `d2.config.js` files, depending on if it's a lib or an app; and libs don't get `App.module.css` or `App.test.jsx`) — this would make the pathing a little easier, and avoids duplicating files like `App.modules.css`, `package.json`, and `README.md` across two init directories
2. For the TS/JS files (like `entrypoint.jsx`), since there are no types in the TS versions yet, it might be fine to keep one init file around (e.g. just the `.jsx` version, but could be either), then give it a TS extension when copying it into a new project if the TS flag is set. If the .ts(x) files had types in them, I think they could be their own files
3. Don’t forget docs 😊 

I'll add a few other things to specific files 👍

cli/config/d2ConfigDefaults.typescript.js
```diff
@@ -0,0 +1,68 @@
+const defaultsApp = {
```
KaiVandivier, line 1: I don't think this file is needed -- these are more JS code to import as defaults, rather than boilerplate to be copied into a new app

(I actually meant to add a comment about this when separating these objects from the boilerplate files, but forgot to add it in the last PR, so it's in a more recent one: [d2ConfigDefaults.js](https://github.com/dhis2/app-platform/pull/867/files#diff-be9857ee2a9ff24deebd7b2ba307a19368a54f46792f7cb1f735fcea95dfe6de))

cli/src/commands/init.js
```diff
@@ -69,22 +69,25 @@ const writeGitignore = (gitignoreFile, sections) => {
...
 }
 
-const handler = async ({ force, name, cwd, lib }) => {
+const handler = async ({ force, name, cwd, lib, typeScript }) => {
+    // create the folder where the template will be generated
     cwd = cwd || process.cwd()
     cwd = path.join(cwd, name)
     fs.mkdirpSync(cwd)
-    const paths = makePaths(cwd)
+    const paths = makePaths(cwd, { typeScript })
```
KaiVandivier, line 77: [Random brainstorming that I think is out of scope for this PR:] It would be cool if passing the `ts` flag here wasn't necessary, but I think that would need a refactoring of the paths to use something like like globs, which may end up being more complex

cli/src/commands/init.js
```diff
@@ -183,7 +186,42 @@ const handler = async ({ force, name, cwd, lib }) => {
...
+
+        // Todo: check why d2 start is renaming TS files to JS
+
+        // ToDO: make custom eslint config part of the template (and copy it)0
+        // similar to: https://github.com/dhis2/data-exchange-app/pull/79/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5
+        // install dependencies needed for eslint
+        // "@typescript-eslint/eslint-plugin"
+        // "@typescript-eslint/parser"
+
+        // ToDO: install any other TS dependencies needed
```
KaiVandivier, line 210: Would this include `types/react[-dom]`?

cli/src/commands/init.js
```diff
@@ -242,6 +280,12 @@ const command = {
             type: 'boolean',
             default: false,
         },
+        typeScript: {
```
KaiVandivier, line 283: _Suuuper_ subjective and equally minor since the alias is set up, and maybe I’m crazy, but I’ll share in case it’s in the back of your mind too: `—typeScript` doesn’t feel right to me as a flag or variable name 😅 I think of it as `typescript` or `ts` — it seems like you used `—typescript` too when you wrote out CLI command script for your workflow

cli/src/commands/init.js
KaiVandivier: Maybe even `@types/jest`

Screenshot 2024-08-05 at 5 31 30 PM

cli/src/commands/init.js
```diff
@@ -183,7 +186,42 @@ const handler = async ({ force, name, cwd, lib }) => {
...
+
+        reporter.info('install TypeScript as a dev dependency')
+        // ToDO: restrict the major version of TS we install
+        await exec({
+            cmd: 'yarn',
+            args: ['add', 'typescript', '--dev'],
+            cwd: paths.base,
+        })
+
+        // Todo: check why d2 start is renaming TS files to JS
```
KaiVandivier, line 202: Is this still happening? I haven't noticed it

Note that some files from the app shell will be copied in to `.d2/shell/`, which are called `App.jsx` and `App.test.jsx`, which may be confusing:

Screenshot 2024-08-05 at 5 32 38 PM

PR #872, opened 2024-08-26
fix: base url env & refactor [LIBS-635]

KaiVandivier:
Part of [LIBS-635](https://dhis2.atlassian.net/browse/LIBS-635)

I noticed that the PUBLIC_URL and BASE_URL env vars weren't making it to built apps after some Vite redesigns -- the cause was that they were added onto `process.env` directly separate from other env var handling, which ended up happening after `process.env` was parsed and filtered for sending to the shell.

An easy fix would be to just move the `process.env` modification _before_ where the shell env gets parsed, but much of the env handling was getting dispersed in several locations, and I thought this could be a good time to try to simplify things.

By making a standalone `getEnv` function, then `bootstrapShell` could be its own function, and then we don't need to keep track of a `shell` object any more, simplifying that concept in the script a bit 👍 

I ended up leaving the baseUrl and publicUrl logic where it is for printing purposes, but tracked them as regular variables and only adding them to the env later with everything else


amcgee (approved):
Much cleaner!  Code looks good ✅ 

Looks like there are just a few leftover comments to clean up?

PR #875, opened 2024-09-02
feat: filter build logs to summarize moment locales

KaiVandivier:
Part of [LIBS-635](https://dhis2.atlassian.net/browse/LIBS-635)

Moment locales can end up being 135 lines of the build output -- this uses a custom logger for Vite to filter out those lines from the build output, and instead print a static summary of the moment locales. My assumption is that the moment locales won't really change, so the static summary is okay

It feels a little weird to override the Vite logger just to make this happen, so this is just a draft for now -- curious about input

If you're curious about how Vite does its logging, it looks like it [uses a Rollup plugin](https://github.com/vitejs/vite/blob/dc3c14f39fcd57867c8ae81b75ae768b53b4b880/packages/vite/src/node/plugins/reporter.ts) to asses the build output and log it

Screenshot 2024-09-02 at 10 33 41 PM

### DIFF
--- a/cli/config/makeViteConfig.mjs
+++ b/cli/config/makeViteConfig.mjs
@@ -1,8 +1,10 @@
+import { chalk } from '@dhis2/cli-helpers-engine'
 import react from '@vitejs/plugin-react'
 import {
     defineConfig,
     searchForWorkspaceRoot,
     transformWithEsbuild,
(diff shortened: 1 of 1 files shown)


KaiVandivier:
Decided against this to keep config simpler; can revisit in the future

PR #878, opened 2024-10-02
feat: migrate snapshot files too [LIBS-681]

KaiVandivier:
[LIBS-681](https://dhis2.atlassian.net/browse/LIBS-681)

Tested out in the Dashboard app, and the snapshots are indeed use for comparison -- the platform changes seem to have removed white space in some snaps as seen in the second screenshot below, for example.

One potential shortcoming of the script I noticed, however, is that it doesn't update file paths for things like mocked jest modules, resulting in errors like:

```sh
 FAIL  src/components/Item/VisualizationItem/__tests__/Item.spec.jsx
  ● Test suite failed to run

    Cannot find module '../../ItemHeader/DeleteItemButton.js' from 'src/components/Item/VisualizationItem/__tests__/Item.spec.jsx'
```

### Screenshots

snap-migration

snap-test

[LIBS-681]: https://dhis2.atlassian.net/browse/LIBS-681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/src/commands/migrate/js-to-jsx.js
+++ b/cli/src/commands/migrate/js-to-jsx.js
@@ -34,8 +34,9 @@ const isJsxInFile = async (filepath) => {
     }
 }
 
+const jsRegex = /\.js(\.snap)?/ // handles .js or .js.snap files
 const renameFile = async (filepath) => {
-    const newPath = filepath.concat('x') // Add 'x' to the end to make it 'jsx'
+    const newPath = filepath.replace(jsRegex, '.jsx$1')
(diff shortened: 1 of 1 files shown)


kabaros (approved):
LGTM - will you add a ticket  for updating jest mocks? or we can just tell people to manually update (it's easily spotted from the error)

amcgee (approved):

PR #881, opened 2024-10-07
fix: return result of parent alerts add [LIBS-695]

KaiVandivier:
Part of [LIBS-695](https://dhis2.atlassian.net/browse/LIBS-695)

Returns result of `parentAlertsAdd` (an alert) for `useAlert()` in the plugin context to receive

See also https://github.com/dhis2/app-runtime/pull/1388 -- it would probably make sense to merge that one first, then update the app-runtime dep in platform 👍 

- [x] Update app-runtime dep here
- [x] Catch up with `alpha`

[LIBS-695]: https://dhis2.atlassian.net/browse/LIBS-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amcgee (approved):
Nice find!!

PR #882, opened 2024-10-08
fix: bump ui library

kabaros:
to avoid all the propTypes errors in React 18


amcgee (approved):
LGTM

❓ should we also update the peerDep here?  https://github.com/dhis2/app-platform/blob/master/adapter/package.json#L43

PR #885, opened 2024-10-18
fix: bump typescript version used in app shell

kabaros:


kabaros:
there is a separate discussion on whether this dependency is necessary in the first place - for the time being, it needs to be bumped at least, since it causes conflicts with the changes on cli-style to introduce TypeScript linting

amcgee (approved):

PR #888, opened 2024-10-25
fix: remove custom eslint from TS template

kabaros:
now that eslint for TS is integrated in cli-style, no need to do anything special for linting eslint


kabaros:
consumers can install eslint like they do in the current JS template with `d2 style add eslint` - although I think it'd be a good idea to add these by default to both JS and TS templates?

amcgee (approved):

PR #889, opened 2024-10-28
fix: env refactor for Vite wrap-up [LIBS-690]

KaiVandivier:
Part of [LIBS-690](https://dhis2.atlassian.net/browse/LIBS-690)

### `process.env` vs. `import.meta.env`

Vite adds environment variables to the `import.meta.env` namespace by default. My plan was to make this the standard for the platform, too, because I thought this was a standard format that goes along with ES Modules. It turns out that it’s not standard, actually; it’s a pattern specific to Vite. Other build tools tend to use `process.env` as the default.

I've noticed that there are some other tools and packages out there that depend on configuration variables that are set on `process.env`, so dropping support for it entirely in favor of `import.meta.env` may be a bit weird.

This makes me think we should continue adding env vars to `process.env`, since it’s a little more generic, instead of migrating entirely to `import.meta.env` as the main env var namespace for the platform.

We can still deprecate the `REACT_APP_`-prefixed variable names, i.e. `process.env.REACT_APP_DHIS2_APP_VERSION` => `process.env.DHIS2_APP_VERSION`. We can handle it in the same way we planned to handle moving to `import.meta.env` i.e., have both variable names available in v12, then remove the `REACT_APP_` versions in v13.

If we do, I’m not sure how much attention to give `import.meta.env` — Right now, Vite is configured to filter the env for values starting with “DHIS2_” and add them to `import.meta.env`. It seems that only values on `import.meta.env` can be used in HTML files, like `%DHIS2_APP_NAME%` in `index.html`. I propose we kinda keep `import.meta.env` as-is; some people may want to use it, and it doesn't really hurt to add the `DHIS2_` prefixed env vars onto it.

### Other env vars

I made some other changes to make the env var objects more concise:
1. Remove PWA default variables, and instead handle `undefined` values
2. Remove variables with `undefined` values from the env

I tested these out by starting and building both the simple-app and pwa-app examples to make sure they work with and without env vars. I also inspected the SW's caching behavior in the PWA app, and it looks good.

### Examples

Before:

env-pwa-before-console
env-pwa-before-cli

After:

env-pwa-after-console
env-change-after-cli

[LIBS-690]: https://dhis2.atlassian.net/browse/LIBS-690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amcgee (approved):
Code LGTM

PR #894, opened 2024-11-21
fix: clear only required build dirs

amcgee:
Instead of clobbering all of `./build`, just remove `./build/app` and `./build/bundle` before writing build output there.  The library compiler (`src/lib/compile.js`) already clears each module-specific output dir (`./build/es`, `./build/cjs`)

In the future we might not want to clobber all the bundle output either... sometimes it can be beneficial to keep different bundle versions.

We could optionally move these up to the beginning of the build process if we want them also the be cleared on unsuccessful builds.

### DIFF
--- a/cli/src/commands/build.js
+++ b/cli/src/commands/build.js
@@ -75,8 +75,6 @@ const handler = async ({
         ? getAppParameters(standalone, config)
         : null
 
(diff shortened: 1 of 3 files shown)


KaiVandivier (approved):
LGTM 👍

PR #906, opened 2025-02-06
feat(vite): add source maps to build

KaiVandivier:
This may be an expected default, but also makes it possible to use [Teamscale](https://docs.teamscale.com/howto/setting-up-profiler-tga/javascript/) to collect test coverage of Vite apps easily using a standalone build

(CRA added source maps by default)


amcgee (approved):
LGTM

PR #908, opened 2025-02-28
fix: surface error when 'd2-app-scripts test' fails

kabaros:
This surfaces the error that happens when running `yarn test`, otherwise the error is swallowed and we just get an unhelpful generic error message.

| Before | After |
| -- | -- |
| image | image |


d-rita (approved):
Quite useful. LGTM!

PR #914, opened 2025-03-14
feat: add typescript hints for d2 config

kabaros:
this PR adds type definition and documentation for d2config, which then consumers can use to get type hints in their d2config even if they're not using TypeScript

NOTE: the video is outdated, now we need to import the config declaration from cli-app-scripts rather than app-runtime  `/** @type {import('@dhis2/cli-app-scripts').D2Config} */`

[d2config-types.webm](https://github.com/user-attachments/assets/bda03306-f4de-4a2c-bdf3-33f3d0b0c154)


KaiVandivier (approved):
Awesome! 🙌

PR #915, opened 2025-03-19
fix(pwa): ensure service worker registration triggers on page load

KaiVandivier:
If `register()` is called after the window `load` event, then the service worker won't get registered. This change checks for the [document ready state](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) to check if it's loaded already


amcgee (approved):
Tested and works well ✅

PR #919, opened 2025-05-12
ci: update caching method

KaiVandivier:


.github/workflows/dhis2-verify-commits.yml
```diff
@@ -9,7 +9,10 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v2
-            - uses: c-hive/gha-yarn-cache@v1
+            - uses: actions/setup-node@v3
+              with:
+                  node-version: 18.x
+                  cache: 'yarn'
```
KaiVandivier, line 15: This change is a little different than from similar ones, since it _adds_ `actions/setup-node` here -- I'm guessing that's no problem, but just want to make sure

kabaros (approved):

PR #932, opened 2025-07-14
feat: allow custom Vite config extensions

KaiVandivier:
[LIBS-706](https://dhis2.atlassian.net/browse/LIBS-706)


kabaros (approved):
code LGTM and I've tested this using the example provided, but I don't have any other use cases in mind (let me know if there are other specific things to test)

PR #933, opened 2025-07-23
feat: decouple cli-app-scripts from yarn

kabaros:


cli/src/lib/bootstrapShell.js
```diff
@@ -66,14 +66,17 @@ const bootstrapShell = async ({ paths, shell, force = false }) => {
...
+    )
     if (fs.exists(srcNodeModules)) {
         reporter.debug(
             `Linking ${path.relative(
                 paths.base,
                 destNodeModules
             )} to ${path.relative(paths.base, srcNodeModules)}...`
         )
-        await fs.ensureSymlink(srcNodeModules, destNodeModules)
+        fs.ensureSymlinkSync(srcNodeModules, destNodeModules)
```
kabaros, line 79: this - and other fs.sync calls - fix a race condition that happened before the first time we run `start` where the symlink is sometimes not yet created

cli/src/commands/init.js
```diff
@@ -69,14 +69,31 @@ const writeGitignore = (gitignoreFile, sections) => {
     fs.writeFileSync(gitignoreFile, gitignore.stringify(sections, format))
 }
 
-const handler = async ({ force, name, cwd, lib, typeScript }) => {
+const handler = async ({ force, pnpm, npm, name, cwd, lib, typeScript }) => {
```
KaiVandivier: Add warning to use `pnpm create @dhis2/app` instead?

cli/src/lib/packageManagers/isNpmProject.js
```diff
@@ -0,0 +1,7 @@
+const paths = require('../paths')
+
+const isNpmProject = () => {
+    return paths.npmLock !== null
+}
+
+module.exports = isNpmProject
```
KaiVandivier: Maybe check for `packageManager` field in `package.json` first?

(This might return an erroneous value if there are multiple lock files in a project, which happens)

PR #962, opened 2026-08-19
fix: stabilize config object

HendrikThePendric:
This stabilises the config object as much as possible, which comes in handy when we pass it down the component tree.

For more context see the description of https://github.com/dhis2/app-runtime/pull/1445


KaiVandivier (approved):
Nice :)

